PR #10, opened 2018-06-17
Fix missing hcl tag in floating ip docs

arminbuerkle:
Also lower case 'servers' where appropriate.


thetechnick:
Thank you for the fix.

PR #15, opened 2018-08-01
Reverse dns configuration

### LINKED ISSUE #5: Reverse DNS configuration
Moved from https://github.com/hetznercloud/terraform-provider-hcloud/issues/32

LKaemmerling:
This should fix #5 and make the reverse dns configuration easily editable through a new resource.


hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+}
+
+func resourceReverseDnsRead(d *schema.ResourceData, m interface{}) error {
+	client := m.(*hcloud.Client)
+	ctx := context.Background()
+
+	ipAddress := string(d.Get("ip_address").(string))
+	rdnsType := ""
+	id, ok := d.GetOk("server_id")
```
thetechnick: Please lookup the server/floating ip via the assigned record id.
You could use a function like this:

```go
func lookupRDNSID(id string) (server *hcloud.Server, floatingIP *hcloud.FloatingIP, err error) {
	// lookup code
	return
}
```
and use it like this:
```go
server, floatingIP, err = lookupRDNSID(id)
if err != nil {
	if err == errInvalidRDNSID {
		// delete from state
	}
	return err
}
if server != nil {
	// handle server
}
// floatingIP must be != nil here because otherwise 'lookupRDNSID' would,
// have returned an error "not found" or "invalid id"
```

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,240 @@
...
+			},
+		},
+	}
+}
+
+func resourceReverseDnsRead(d *schema.ResourceData, m interface{}) error {
+	client := m.(*hcloud.Client)
+	ctx := context.Background()
+
+	ipAddress := string(d.Get("ip_address").(string))
```
thetechnick: superfluous cast to string

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,240 @@
...
+			"ip_address": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"dns_ptr": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
```
thetechnick: Why is the `dns_ptr` optional? The the code will break if it is not provided.

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,241 @@
...
+				ForceNew: true,
+			},
+			"floating_ip_id": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				ForceNew: true,
+			},
+			"ip_address": {
+				Type:     schema.TypeString,
+				Required: true,
```
thetechnick: `ip_address` could be made optional by defaulting to the public IPv4 address or the first address of the IPv6 subnet.
What are your thoughts on this?

I have not made up my mind yet, but this seems to be a nice shortcut for the majority use case.

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,241 @@
...
+	rdnsType := ""
+	id, ok := d.GetOk("server_id")
+	if ok == false {
+		id, ok = d.GetOk("floating_ip_id")
+		switch {
+		case ok == false:
+			log.Printf("[WARN] Invalid id (%s), removing from state: %v", d.Id(), ok)
+			d.SetId("")
+			return nil
+		case ok == true:
```
thetechnick: superfluous second `case` after return and check `booleans` directly:
```
ok == true => ok
ok == false => !ok
```

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,241 @@
...
+			rdnsType = "floating_ip"
+		}
+	} else {
+		rdnsType = "server"
+	}
+	ip := net.ParseIP(d.Get("ip_address").(string))
+	if ip == nil {
+		log.Printf("[WARN] The ip you provide (%s) is not valid, removing from state", ipAddress)
```
thetechnick: Please return an error to the user and do not delete the record from state.

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+		if floatingIP == nil {
+			log.Printf("[WARN] Floating IP (%s) not found, removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+
+		switch floatingIP.Type {
+		case "ipv4":
+			if strings.Count(ipAddress, ".") != 3 {
```
thetechnick: Use `net.ParseIP` here and check if the ip is a IPv4 address with https://golang.org/pkg/net/#IP.To4

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,241 @@
...
+				log.Printf("[WARN] Floating IP (%s) is an ipv4 floating ip but you dont write a valid one, removing from state", d.Id())
+				d.SetId("")
+				return nil
+			}
+		case "ipv6":
+			if strings.Count(ipAddress, ":") > 1 {
```
thetechnick: Use `net.ParseIP`

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,241 @@
...
+				log.Printf("[WARN] Floating IP (%s) is an ipv6 but you write an invalid ip, removing from state", d.Id())
+				d.SetId("")
+				return nil
+			}
+		}
+		d.Set("dns_ptr", floatingIP.DNSPtrForIP(ip))
```
thetechnick: `DNSPtrForIP` might return an empty string when the DNSPtr is not found. In this case the rdns entry should be removed from state, because the rdns entry was deleted outside of terraform.

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,241 @@
...
+		d.SetId(fmt.Sprintf("f-%d-"+ipAddress, id))
```
thetechnick: Please create a new function that generates the id for floating ips and servers.

Something like this:
```go
func generateRDNSID(server *hcloud.Server, floatingIP *hcloud.FloatingIP, ip string) string {
	if server != nil {
		return fmt.Sprintf("s-%d-%s", id, ip)
	}
	if floatingIP != nil {
		return fmt.Sprintf("f-%d-%s", id, ip)
	}
	return ""
}
```
Also note that if you use `fmt.Sprintf` you should use it with multiple parameters and not mix it with string concatenation.

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,241 @@
...
+	id, ok := d.GetOk("server_id")
+	if ok == false {
+		id, ok = d.GetOk("floating_ip_id")
+		switch {
+		case ok == false:
+			log.Printf("[WARN] Invalid id (%s), removing from state: %v", d.Id(), ok)
+			d.SetId("")
+			return nil
+		}
+		d.SetId(fmt.Sprintf("f-%d-"+ip, id))
```
thetechnick: Set the record id *after* checking if the config is valid and the server/floating ip exists.

thetechnick (changes requested):

hcloud/resource_hcloud_reversedns.go
LKaemmerling: It is optional because we need to make it possible to "change to the default" value

thetechnick: When you want to change to the default value you have to delete the terraform resource, this will reset/delete the reverse dns pointer.
Using terraform for negative assertions for your infrastructure like: "make sure this dns ptr is NOT set" does not make much sense to me.

hcloud/resource_hcloud_reversedns.go
thetechnick: scratch this idea
Defaulting can only work with a single IP address or we cannot create a meaningful unique id for the terraform record. And for a only the ipv4, I doubt it is worth the hassle.

We can still add defaulting later, because it would not break existing configs.

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,226 @@
...
+			"ip_address": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"dns_ptr": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
```
thetechnick: `ForceNew` should not be needed for the `dns_ptr`, because it would trigger a `delete` for the resource before recreating it. When only the `dns_ptr` has changed, the rnds entry can be updated without having to delete it first.

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,226 @@
...
+			},
+		},
+	}
+}
+
+func resourceReverseDNSRead(d *schema.ResourceData, m interface{}) error {
+	client := m.(*hcloud.Client)
+	ipAddress := string(d.Get("ip_address").(string))
```
thetechnick: superfluous cast to string

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,226 @@
...
+	server, floatingIP, err := lookupRDNSID(d.Id(), client)
+
+	switch {
+	case err != nil:
```
thetechnick: use `if` instead

If there is any error with the lookup process now, the record is deleted without the user ever knowing why.
You should create a new "Invalid id" error and check if it is returned before deleting the record from terraform.
Other errors should be returned directly so the terraform step fails.

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,226 @@
...
+		log.Printf("[WARN] Invalid id (%s), removing from state: %s", d.Id(), err)
+		d.SetId("")
+		return nil
+	}
+	ip := net.ParseIP(d.Get("ip_address").(string))
+	if ip == nil {
+		log.Printf("[ERR] The ip you provide (%s) is not valid", ipAddress)
+		return nil
```
thetechnick: return the error instead of logging it, to inform the user.

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	switch {
+	case floatingIP != nil:
+		if floatingIP == nil {
```
thetechnick: unreachable code

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,226 @@
...
+				d.SetId("")
+				return nil
+			}
+		}
+		dnsPtr := floatingIP.DNSPtrForIP(ip)
+		if dnsPtr != "" {
+			d.Set("dnsPtr", dnsPtr)
+			d.SetId(generateRDNSID(nil, floatingIP, ipAddress))
+		} else {
+			d.SetId("")
```
thetechnick: Please log when you remove an item from state in the read function

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	switch {
+	case floatingIP != nil:
+		if floatingIP == nil {
+			log.Printf("[WARN] Floating IP (%s) not found, removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+
+		ip = net.ParseIP(ipAddress)
```
thetechnick: `net.ParseIP` might return `nil` if the ip address has an invalid format

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,226 @@
...
+		switch floatingIP.Type {
+		case "ipv4":
+			if ip.To4() == nil {
+				log.Printf("[WARN] Floating IP (%s) is an ipv4 floating ip but you dont write a valid one, removing from state", d.Id())
```
thetechnick: Please rephrase this log message.

The `ip_address` value would be read from the `.tfstate` file so it is not directly provided by the user. Assuming the code in `create` works correct this case can never happen, so it is just a worse case "user edited the state file manually" or there is a big screw up in `create` or `update`

**important**
You do not have to parse the ip addresses here at all, because if the ip address is not found you are removing the item from the state anyway.

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,226 @@
...
+				d.SetId("")
+				return nil
+			}
+		case "ipv6":
+			if ip.To16() == nil {
+				log.Printf("[WARN] Floating IP (%s) is an ipv6 but you write an invalid ip, removing from state", d.Id())
```
thetechnick: see above.

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,226 @@
...
+				d.SetId("")
+				return nil
+			}
+		}
+		dnsPtr := floatingIP.DNSPtrForIP(ip)
+		if dnsPtr != "" {
+			d.Set("dnsPtr", dnsPtr)
```
thetechnick: You also have to `.Set()` all other `non ID` properties otherwise `terraform import` can not work.

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,226 @@
...
+	case server != nil:
+		ip = net.ParseIP(ipAddress)
+		if ip.To4() != nil {
+			d.SetId(generateRDNSID(server, nil, ipAddress))
+			d.Set("dnsPtr", server.PublicNet.IPv4.DNSPtr)
+		} else if ip.To16() != nil {
+			for rdns := range server.PublicNet.IPv6.DNSPtr {
+				if rdns == ipAddress {
+					d.SetId(generateRDNSID(server, nil, ipAddress))
+					d.Set("dnsPtr", server.PublicNet.IPv6.DNSPtrForIP(ip))
```
thetechnick: Set all properties like above in the floating ip case.

hcloud/resource_hcloud_reversedns.go
LKaemmerling: Yeah, but then we need a update method or?

thetechnick: There should always be a update method for terraform resources

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,208 @@
...
+
+	id, ok := d.GetOk("server_id")
+	if ok == false {
+		id, ok = d.GetOk("floating_ip_id")
+		switch {
+		case !ok:
+			log.Printf("[WARN] Invalid id (%s), removing from state: %v", d.Id(), ok)
+			d.SetId("")
+			return nil
+		}
```
thetechnick: use if instead

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,208 @@
...
+		floatingIP, _, _ := client.FloatingIP.GetByID(ctx, id.(int))
+		d.SetId(generateRDNSID(nil, floatingIP, ip))
+		rdnsType = "floatingIP"
+	} else {
+		server, _, _ := client.Server.GetByID(ctx, id.(int))
+		d.SetId(generateRDNSID(server, nil, ip))
+		rdnsType = "server"
+	}
+
+	switch {
```
thetechnick:
```go
switch rdnsType {
  case "floating_ip":
  case "server":
}
```

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,208 @@
...
+		floatingIP, _, err := client.FloatingIP.GetByID(ctx, id.(int))
+		if err != nil {
+			return err
+		}
+		if floatingIP == nil {
+			log.Printf("[WARN] Floating IP (%s) not found, removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+		action, _, err := client.FloatingIP.ChangeDNSPtr(ctx, floatingIP, ip, &ptr)
```
thetechnick: err is never checked

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,208 @@
...
+		server, _, err := client.Server.GetByID(ctx, id.(int))
+		if err != nil {
+			return err
+		}
+		if server == nil {
+			log.Printf("[WARN] Server (%s) not found, removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+		action, _, err := client.Server.ChangeDNSPtr(ctx, server, ip, &ptr)
```
thetechnick: err is never checked

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,208 @@
...
+	if floatingIP != nil {
+		return fmt.Sprintf("f-%d-%s", floatingIP.ID, ip)
+	}
+	return ""
+}
+func lookupRDNSID(id string, client *hcloud.Client) (server *hcloud.Server, floatingIP *hcloud.FloatingIP, err error) {
+
+	ctx := context.Background()
+	ids := strings.Split(id, "-")
+	_id, _ := strconv.Atoi(string(ids[1]))
```
thetechnick: superfluous cast to string

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,208 @@
...
+	switch {
+	case err != nil:
+		log.Printf("[WARN] Invalid id (%s), removing from state: %s", d.Id(), err)
+		d.SetId("")
+		return nil
+	}
+	if floatingIP != nil {
+		if d.HasChange("dns_ptr") {
+			ip := d.Get("ip_address").(string)
+			action, _, _ := client.FloatingIP.ChangeDNSPtr(ctx, floatingIP, ip, nil)
```
thetechnick: error is ignored

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,208 @@
...
+
+			if err := waitForFloatingIPAction(ctx, client, action, floatingIP); err != nil {
+				return err
+			}
+		}
+	} else if server != nil {
+		ip := d.Get("ip_address").(string)
+		action, _, _ := client.Server.ChangeDNSPtr(ctx, server, ip, nil)
```
thetechnick: error is ignored

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,208 @@
...
+func resourceReverseDNSDelete(d *schema.ResourceData, m interface{}) error {
+	client := m.(*hcloud.Client)
+	ctx := context.Background()
+	server, floatingIP, err := lookupRDNSID(d.Id(), client)
+	switch {
+	case err != nil:
+		log.Printf("[WARN] Invalid id (%s), removing from state: %s", d.Id(), err)
+		d.SetId("")
+		return nil
+	}
```
thetechnick: use an `if`, no single case switches ;)

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,262 @@
...
+
+	id, ok := d.GetOk("server_id")
+	if ok == false {
+		id, ok = d.GetOk("floating_ip_id")
+		if !ok {
+			log.Printf("[WARN] Invalid id (%s), removing from state: %v", d.Id(), ok)
+			d.SetId("")
+			return nil
+		}
+		floatingIP, _, _ := client.FloatingIP.GetByID(ctx, id.(int))
```
thetechnick: err is ignored

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,262 @@
...
+		d.SetId(generateRDNSID(nil, floatingIP, ip))
+		rdnsType = "floatingIP"
+	} else {
+		server, _, _ := client.Server.GetByID(ctx, id.(int))
```
thetechnick: err is ignored

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,262 @@
...
+		d.SetId(generateRDNSID(server, nil, ip))
+		rdnsType = "server"
+	}
+
+	switch rdnsType {
+	case "floating_ip":
```
thetechnick: Switch case `"floating_ip"` differs from `rdnsType = "floatingIP"` in line 112

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	ip := d.Get("ip_address").(string)
+	ptr := d.Get("dns_ptr").(string)
+
+	if floatingIP != nil {
+		if d.HasChange("dns_ptr") {
+			if err != nil {
+				return err
+			}
+			if floatingIP == nil {
```
thetechnick: code never executes:
```go
if floatingIP != nil {
  if floatingIP == nil {
  }
}
```

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,262 @@
...
+			}
+			if err := waitForFloatingIPAction(ctx, client, action, floatingIP); err != nil {
+				return err
+			}
+		}
+	} else if server != nil {
+		if err != nil {
+			return err
+		}
+		if server == nil {
```
thetechnick: code never executes:
```go
if server != nil {
  if server == nil {
  }
}
```

hcloud/resource_hcloud_reversedns.go
```diff
@@ -0,0 +1,262 @@
...
+
+	}
+	return resourceReverseDNSRead(d, m)
+}
+func resourceReverseDNSDelete(d *schema.ResourceData, m interface{}) error {
+	client := m.(*hcloud.Client)
+	ctx := context.Background()
+	server, floatingIP, err := lookupRDNSID(d.Id(), client)
+	if err != nil {
+		log.Printf("[WARN] Invalid id (%s), removing from state: %s", d.Id(), err)
```
thetechnick: If `lookupRDNSID` returns an error you have cover two cases:
1. Invalid id => log invalid id, remove from state and return nil
2. every other error => directly return the error

Please also make sure the errors of `lookupRNDSID` are handled the same way everywhere the function is used. (line 161)

hcloud/resource_hcloud_rdns.go
```diff
@@ -0,0 +1,310 @@
...
+
+	if floatingIP != nil {
+		dnsPtr := floatingIP.DNSPtrForIP(ip)
+		if dnsPtr == "" {
+			log.Printf("[WARN] RDNS entry (%s) not found, removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+
+		d.Set("dnsPtr", dnsPtr)
```
thcyron: Should be `dns_ptr`.

hcloud/resource_hcloud_server.go
```diff
@@ -82,7 +82,7 @@ func resourceServer() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
-			"ipv6_address": {
+			"ipv6_network": {
```
thcyron: Isn’t this a breaking change?

LKaemmerling: Yeah this is a breaking change, Nico and I discussed this and we decided to introduce this breaking change since the current behavior is not really good and usable.

thetechnick (approved):

Issue #42, opened 2018-11-07
Consider adding volume association resource


MariusRumpf:
Similar to the `hcloud_floating_ip_assignment` it would be nice if one might assign an existing volume to a `hcloud_server` resource the same way. It might look like the following then: 

``` 
resource "hcloud_volume_assignment" "node_1_volume_1" {
    volume_id = "${data.hcloud_volume.volume1.id}"
    server_id = "${hcloud_server.node1.id}"
}

data "hcloud_volume" "volume1" {
    name = "volume1"
}

resource "hcloud_server" "node1" {
  name = "node1"
  image = "debian-9"
  server_type = "cx11"
}
```

LKaemmerling:
Good idea. I will implement it.

LKaemmerling:
Will be released with the next version.

PR #47, opened 2018-11-13
Add volume_attachment resource

LKaemmerling:
Issue: #42 

With this new resource it is possible to attach a volume to a server, after the volume is created. 

@thcyron can you please review this, when you have time?

### DIFF
--- a/hcloud/resource_hcloud_volume_attachment.go
+++ b/hcloud/resource_hcloud_volume_attachment.go
@@ -87,7 +87,7 @@ func resourceVolumeAttachmentRead(d *schema.ResourceData, m interface{}) error {
 		return nil
 	}
 
-	// check if correct server is attached
+	// check if volume is attached to any server
 	if volume.Server == nil {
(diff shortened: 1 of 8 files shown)


hcloud/resource_hcloud_volume_attachment.go
```diff
@@ -0,0 +1,129 @@
...
+		return err
+	}
+	if server == nil {
+		log.Printf("[WARN] Server ID (%v) not found, removing volume attachment from state", d.Get("server_id"))
+		d.SetId("")
+		return nil
+	}
+
+	// check if correct server is attached
+	if volume.Server == nil {
```
thcyron, line 91: Was just checking for nil your intention? The comment sounds like you also wanted to compare IDs.

thcyron (approved):

hcloud/resource_hcloud_volume_attachment.go
LKaemmerling: I changed the comment to be more clear.

PR #52, opened 2018-11-27
add most_recent option to images datasource

nferch:


LKaemmerling:
First of all, thank you for your contribution!

From the code side it looks great! Thank you! I'm not really sure if this is the best solution, because what is if you want the oldest image? I think something like an optional string with the values "latest" or "oldest" would be better. If the string is empty or not available it would throw the error like it is currently implemented.  What do you think about this @nferch ?

nferch:
@LKaemmerling My approach here was to take the simplest possible approach to satisfy my use case, which was to choose the most recent snapshot with a particular label. See https://github.com/hashicorp/packer/pull/7046 for more context.

I didn't envision a scenario where I'd want the oldest image, and wanted to avoid the complexity of more a sophisticated grammar or options to choose which image that matches a selector without a clear need. If for some reason I needed a specific image for debugging or troubleshooting needs, I'd likely just hardcode it since it would be temporary and off the happy path anyway.

Open to discussion on what that could look like, but for the time being I chose adding the `most_recent` option which follows what is present in the `openstack_images_image_v2` and `aws_ami` data sources.

LKaemmerling:
Yeah, you are right. The `most_recent` flag should be enough.

Can you add a test for it?

nferch:
@LKaemmerling sure thing, good idea!

nferch:
whoops, didn't mean to commit that, was just from my testing

LKaemmerling:
Great work, thank you!

PR #106, opened 2019-07-26
Bugfix for private isos

MarkusFreitag:
According to the documentation, ISOs can currently only be specified by their name. As private ISOs do not have a name, the server ressource should support ID/name.

Unfortunately I could not create an acceptance testcase with a private ISO, as you can not upload it via the API.


LKaemmerling (approved):

LKaemmerling:
Looks good, thank you!

Issue #127, opened 2019-10-01
Network ip_range update


fubbyb:
Currently changing the `ip_range` attribute of the `hcloud_network` has no effect.

**Steps to reproduce:**

1. Create network and subnet resources using following code:
```hcl
resource "hcloud_network" "lan" {
  name     = "lan"
  ip_range = "192.168.0.0/16"
}

resource "hcloud_network_subnet" "lan" {
  network_id   = "${hcloud_network.lan.id}"
  type         = "server"
  network_zone = "eu-central"
  ip_range     = "192.168.0.0/24"
}
```

2. Let's change network and subnet `ip_range` attributes:
```hcl
resource "hcloud_network" "lan" {
  name     = "lan"
  ip_range = "10.0.0.0/8"
}

resource "hcloud_network_subnet" "lan" {
  network_id   = "${hcloud_network.lan.id}"
  type         = "server"
  network_zone = "eu-central"
  ip_range     = "10.0.0.0/16"
}
```

3. Terraform plan:
```
Terraform will perform the following actions:

  # hcloud_network.lan will be updated in-place
  ~ resource "hcloud_network" "lan" {
        id       = "***"
      ~ ip_range = "192.168.0.0/16" -> "10.0.0.0/8"
        labels   = {}
        name     = "lan"
    }

  # hcloud_network_subnet.lan must be replaced
-/+ resource "hcloud_network_subnet" "lan" {
      ~ gateway      = "192.168.0.1" -> (known after apply)
      ~ id           = "***-192.168.0.0/24" -> (known after apply)
      ~ ip_range     = "192.168.0.0/24" -> "10.0.0.0/16" # forces replacement
        network_id   = ***
        network_zone = "eu-central"
        type         = "server"
    }

Plan: 1 to add, 1 to change, 1 to destroy.
```

Execution throws following error:

```
Error: invalid IP range (must be part of network and not overlap with existing ranges) (service_error)
```

**Possible solution**

Force resource recreation on changing `ip_range` of the network.

LKaemmerling:
This is not possible. Please see our documentation for this:
https://docs.hetzner.cloud/#network-actions-change-ip-range-of-a-network

```
[...]

PR #128, opened 2019-10-02
Force network recreation on cidr changing

fubbyb:


fubbyb:
@LKaemmerling This PR related to our discussion in #127. Let's ask Terraform to remove and create this network resource for us.

LKaemmerling:
This is not correct, not every change to the ip_range needs a force new creation of this. Even when you recreate the network all servers will need to be assigned to the new network and will get a new IP. So this behavior is fully that what we want. If you need to change the IP Range to another IP range instead of increasing the given range, you should simply delete the old network and create a new one.

apricote:
Using [`CustomizeDiff`](https://www.terraform.io/docs/extend/resources/customizing-differences.html) and `customdiff.ForceNewIfChange`, it should be possible to check if the new `ip_range` is an extension/superset of the previous range, and only force the recreation if it is a completly different range.

fubbyb:
The Go client implementation for HCloud doesn't have "ip_range" field in network update function. So currently you can not use a Customdiff.

When the client developers add this functionality, then it will be possible to return to the implementation of updating the "ip_range" attribute of the network resource.

In our company, we use Gitops flow, and now we had to solve this issue in two commits:
- remove the network resource
- add network resource with new cidr

Therefore, the downtime was a few minutes, not a few seconds.

LKaemmerling:
@fubbyb The problem on this is that you didn't choose the correct private IP Range at the beginning. It is not possible to do this without the deleting of the network resource, because all servers need to be unassigned, then delete the network. This should be an explicit process because it changes the IP(s) from the servers too.

vintury:
Hello.
I agree with @fubbyb - "If we cannot change the resource, then we can delete it and create a new one."

@LKaemmerling, you can detach network from hosts before remove?

I have big infrastructure.
Any error with the subnet mask will lead to manual intervention in the terraform database. This is unacceptable, as it can lead to complete inaccessibility of the infrastructure!

PR #134, opened 2019-11-11
resources/hcloud_server: Add retry mechanism for enabling the rescue mode

LKaemmerling:
Sometimes the enabling of the rescue mode fails because of a host-side behavior. After talking to Jonas, we should just retry in this case.


thcyron:
LGTM

PR #152, opened 2020-02-15
Adjust Image Name Attribute Documentation

dimisjim:
provider 1.15 outputs this error: "Error: "image" must have more then 0 characters. Have you set the name instead of an ID?" when name attribute is used. By using id, it works fine.

### DIFF
--- a/website/docs/d/image.html.md
+++ b/website/docs/d/image.html.md
@@ -21,7 +21,7 @@ data "hcloud_image" "image_3" {
 }
 
 resource "hcloud_server" "main" {
-  image  = "${data.hcloud_image.image_1.name}"
+  image  = "${data.hcloud_image.image_1.id}"
 }
 ```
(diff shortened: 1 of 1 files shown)


LKaemmerling:
Did you try to use a snapshot/backup or a system image? Snapshots and Backups do not have a `name`, system images have `names` (like:  `ubuntu-18.04`)

dimisjim:
tried to use an image built by packer (sourced by a data source) in hcloud_server's image argument.

LKaemmerling:
Yep, this is than a snapshot. So you need to use the ID then, i would favor adding a sentence to the docs which reflect this better instead of changing only the sample.

dimisjim:
Ok, didn't know there was a difference.

Feel free to change this PR accordingly :)

LKaemmerling:
Thank you!

PR #164, opened 2020-05-14
Argument server does not exists

xsak:
The argument server doesn't exist. The correct format could be found in the example.


LKaemmerling:
Thank you!

PR #186, opened 2020-07-08
Determine default listen port based on protocol

fhofherr:
As discovered by @vmozgovoy in PR #185 the listen port needs to be
derived from the protocol if it is not set in the Terraform
configuration. This commit extracts @vmozgovoy's solution as PR#185
contains additional changes which are already fixed in hcloud-go
v1.18.2.


fhofherr:
The test failures are expected. Once #187 is fixed an this PR is rebased the tests should be green again.

LKaemmerling:
Making this change shouldn't be necessary, as within the Read method, the listen port is already set.

PR #237, opened 2020-09-07
Updating module path to reflect changing organization

stack72:
Without this, anyone using the module directly will have
issues because the module path does not represent the
organisation it sits in

this is in line with other providers that have moved out
of the terraform-providers organisation

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/terraform-providers/terraform-provider-hcloud
+module github.com/hetznercloud/terraform-provider-hcloud
 
 require (
 	github.com/hashicorp/go-multierror v1.1.0 // indirect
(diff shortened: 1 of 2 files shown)


LKaemmerling:
@stack72 First of all thank you for this MR. We actually don't recommend using the driver as a module. We also plan to migrate all logic into internal packages, so it won't be usable anymore. The SemVer promise we give is only for the interface between Terraform and the module, not for the module/code, so you can expect breaking changes within minor releases. If you need a stable interface i would recommend using the hcloud-go: https://github.com/hetznercloud/hcloud-go

stack72:
Hi @LKaemmerling 

We use the module for [pulumi-hcloud](https://github.com/pulumi/pulumi-hcloud/) - which uses the schema and CRUD functionality of the module to allow the users of Pulumi to manage their hcloud infrastructure in a multi-language way. I have been working with one of your marketing team about this module

If you think we should discuss this more, please feel free to mail me offline paul AT pulumi.com and we can discuss

Paul

LKaemmerling:
Hey @stack72,

we would recommend using our hcloud-go for a pulumi provider because we plan to refactor the terraform provider itself and move more (or all code) into internal packages. Could you tell us a bit more about how pulumi interacts with the terraform provider? Does it use the same interface as terraform (GRPC endpoint)? If it is the GRPC endpoint, this is stable and won't change as we just want to move the go code containing the business logic.

Issue #161, opened 2020-04-26
Documentation to automatically registering interface at OS level for private network


sebastiankr:
Hi,
I can successfully associate servers with networks. 
```
resource "hcloud_network" "ens10" {
  name = "ens10"
  ip_range = "10.0.0.0/16"
}
resource "hcloud_network_subnet" "ens10" {
  type = "server"
  network_zone = "eu-central"
  ip_range = "10.0.0.0/24"
  network_id = hcloud_network.ens10.id
}

resource "hcloud_server" "node-0" {
  name        = "node-0"
  server_type = "cpx11"
  datacenter  = "nbg1-dc3"
  image       = "ubuntu-18.04"
  ssh_keys    = var.hcloud_ssh_keys
  user_data = templatefile("${path.module}/../../templates/cloud-config.yml.tpl", {
    docker_gpg : file("${path.module}/../../files/docker_gpg"),
    daemon_json : file("${path.module}/../../files/daemon.json")
  })
}

resource "hcloud_server_network" "ens10" {
  server_id  = hcloud_server.node-0.id
  network_id = var.network_id
  ip         = "10.0.0.2"
}
```
In the hetzner console I can see that the network "ens10" is properly attached to the server. But on the server I can only see loopback and eth0:
```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 96:00:00:4f:ea:d4 brd ff:ff:ff:ff:ff:ff
```

How can I automatically register the network interface "ens10"? How would I do it with cloud-init? Should this not be done by the provider with some sort of an `iface` argument?

danielgrant:
I ran into this same issue last week - I'm using an existing `hcloud_network` that already has a number of servers attached to it.

However, I've gone to add a new server via Terraform, and set it up as above (with a new `hcloud_server_network`), and after the server starts there's no sign of the new private network interface. All the tinkering I could do couldn't bring it up.

In the end, I navigated to the Hetzner Cloud Console, and using the web UI I detached the server from the private network, and then re-added it, and the private network interface appeared!

[...]

PR #264, opened 2020-11-04
resource_hcloud_server_network: sleep before creating a hcloud_server_network resource

flokli:
There seems to be a bug in the Hetzner platform when private networks
are attached immediately after the corresponding server has been
created.

Attaching shortly after, or detaching and reattaching seems to work.

This has been confirmed to be a bug in April at
https://github.com/hetznercloud/terraform-provider-hcloud/issues/161#issuecomment-621125799,
but according to the issue history there, there's still no fix deployed.

Simply sleeping some time before creating the `hcloud_server_network`
resource seems to give the platform enough time to create the machine,
so let's use this as a workaround until the bug in the platform has been
fixed.


LKaemmerling:
Hey @flokli,

thank you! This shouldn't be needed anymore, as we implemented mitigation on the host side.

flokli:
I can confirm the problem doesn't occur anymore. Thanks for fixing this on the host side!

PR #281, opened 2020-12-11
Use a set for inline load balancer targets

fhofherr:
This hopefully makes the state independent of the actual order the
targets were returned by the Hetzner Cloud Backend API.


LKaemmerling:
Confirmed that the change did not break something and fixed the problem. Thank you!

Issue #307, opened 2021-02-03
documentation bug in hcloud-volume


selfscrum:
The description says that `location` is optional when `server_id` is set. On apply, Terraform says that only one of the attributes can be set.

LKaemmerling:
Hey, thank you! This was fixed with #309

PR #309, opened 2021-02-04
Fix hcloud_volume resource: Automount did not work when changing an already existing volume

### LINKED ISSUE #308: Volume automount not working when disk is in same terraform script
I want to create a server and a volume which is automounted to that server.
However, when I provision disk and server in the same script, the disk is attached to the server but not mounted. 

If i create a script where I pass in the id of an already existing disk, volume_attachment works as designed.

```
resource "hcloud_server" "test" {
  name        = var.cluster_name
  image       = var.image
  server_type = var.server_type
  location    = var.location
  labels      = var.labels
  ssh_keys    = [ var.ssh_key ]
  user_data   = var.user_data
}

# tried with these combinations:
# - with location only, and automount in a separate volume_attachment resource
# - with server_id and automount as seen here
# - with depends_on and without
# 
resource "hcloud_volume" "data" {
  name = format("%s-data-disk", var.cluster_name)
  server_id = hcloud_server.test.id
  format = "ext4"
  size = 50
  automount = true
}
```

What's going on?

LKaemmerling:
This bug was reported by a customer in Issue #308. He tried to reassign a volume to another server and then expected the volume to be automounted. Within our `hcloud_volume_attachment` this works, only `hcloud_volume` did not "auto mount" it. 

After this MR we use `AttachWithOpts` (with Automount Flag support) when reassigning a volume instead of `Attach` which does not provide this functionality.  

Closes #308

One more thing I noticed:
`hcloud_volume` is now identically powerful as `hcloud_volume_attachment`. Maybe it makes sense to deprecate `hcloud_volume_attachment`? What do you think?
Signed-off-by: Lukas Kämmerling <lukas.kaemmerling@hetzner-cloud.de>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 ## unreleased
 
 BUG FIXES:
-* `hcloud_volume` datasource: id is now marked as computed to allow more setups where the id is unkown
-* `hcloud_ssh_key` datasource: id is now marked as computed to allow more setups where the id is unkown
-* `hcloud_network` datasource: id is now marked as computed to allow more setups where the id is unkown
-* `hcloud_image` datasource: id is now marked as computed to allow more setups where the id is unkown
-* `hcloud_certificate` datasource: id is now marked as computed to allow more setups where the id is unkown
+* `hcloud_volume` datasource: id is now marked as computed to allow more setups where the id is unknown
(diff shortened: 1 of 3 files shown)


LKaemmerling:
Additionally, this clarifies the documentation of `hcloud_volume`  as found by a customer in #307

fhofherr (approved):

PR #382, opened 2021-06-19
Docs: Mark example "hcloud_token" variable as sensitive

alias-dev:
Prevents the field's values from showing up in CLI output and in Terraform Cloud.

See [Terraform Docs](https://www.terraform.io/docs/extend/best-practices/sensitive-state.html#using-the-sensitive-flag)

Signed-off-by: Alex Andrews <alias-dev@protonmail.com>


LKaemmerling:
Thank you!

PR #384, opened 2021-06-19
Improve error messages from hcloud-go

LKaemmerling:
This improves the err messages for certain error codes returned from the hcloud api. As a sample: invalid_input errors now show which field contains which error. I decided to not implement this in the hcloud-go because we want to have a different outputs based on the specific project where the hcloud-go is used. The output here is optimized to show all errors with the minimal needed space (single log line) and also already in mind to have more of those special cases.

While implementing this i also fixed a smaller bug in the firewall tests and replaced the deprecated scopelint.

Improves/Fixes #384

Signed-off-by: Lukas Kämmerling <lukas.kaemmerling@hetzner-cloud.de>


LKaemmerling:
Before:
```
 Error: invalid input in field 'rules' (invalid_input)
```


After:
```
 Error: invalid input in field 'rules' (invalid_input): [rules.source_ips => [source_ips needs to have at least one entry]]
```

fhofherr (approved):

PR #390, opened 2021-06-25
Update dependencies

fhofherr:
This commit updates the following dependencies:

1. terraform-plugin-sdk to ~2.6.1~. 2.7.0
2. golang.org/x/crypto to v0.0.0-20210616213533-5ff15b29337e

PR #392, opened 2021-06-25
Increase create timeout for servers and snapshots

### LINKED ISSUE #312: 20 min timeout on hcloud_server
Hey team,

We have run into an issue which became a blocker for us. 

We are trying to create "hcloud_server" resources with a pre-built image (snapshot), which sometimes gets a "context deadline exceeded" error after 20 minutes.

In the Hetzner console we can see that these servers are being created, but this takes sometimes 40-60 minutes. When 20 minutes are reached, Terraform reaches an error, and as the error stopped the Terraform apply execution, we have to rerun it, but even then sometimes it fails with the same error, but sometimes it goes through, depending if Hetzner cloud will manage to deploy the servers within 20 min time frame.

We have tested this with small and large images too, tried different instance sizes and non of these made any difference. The only way to make it work fast is to use default Hetzner images (not custom snapshots), but that option is not ideal for our use case, as we have pre-build images ready to be deployed. Our images are fairly large 20-30GB in size if that is relevant.

Could you please tell us if we could increase the 20 minute timeout so at least Terraform would not stop running the apply? This would be an ideal solution. At the moment, we cannot find a way.

Are there any other alternatives?

`module.application-cluster.hcloud_server.ln-app[1]: Still creating... [19m30s elapsed]
module.application-cluster.hcloud_server.ln-app[2]: Still creating... [19m30s elapsed]
module.application-cluster.hcloud_server.ln-app[2]: Still creating... [19m40s elapsed]
module.application-cluster.hcloud_server.ln-app[0]: Still creating... [19m40s elapsed]
module.application-cluster.hcloud_server.ln-app[1]: Still creating... [19m40s elapsed]
module.application-cluster.hcloud_server.ln-app[2]: Still creating... [19m50s elapsed]
module.application-cluster.hcloud_server.ln-app[1]: Still creating... [19m50s elapsed]
module.application-cluster.hcloud_server.ln-app[0]: Still creating... [19m50s elapsed]

Error: context deadline exceeded
Error: context deadline exceeded
Error: context deadline exceeded
ERROR: Job failed: exit code 1`

Thank you
Gabrijel

fhofherr:
In #312 customers reported that server creation can time out if large
snapshots are used. This commit increases the timeout for server
creation to 90 minutes. Since creating snapshots can also take quite
some time the timeout is increased as well.

 Closes #312

### DIFF
--- a/internal/server/resource.go
+++ b/internal/server/resource.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 
@@ -31,7 +32,9 @@ func Resource() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
 		},
-
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(90 * time.Minute),
+		},
 		Schema: map[string]*schema.Schema{
 			"name": {
 				Type:     schema.TypeString,
--- a/internal/snapshot/resource.go
+++ b/internal/snapshot/resource.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"log"
 	"strconv"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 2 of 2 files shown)


fhofherr:
Draft until #390 is merged

LKaemmerling (approved):

PR #419, opened 2021-07-29
Update hcloud_firewall documentation on how to define port range

### LINKED ISSUE #418: [Bug]: documentation, hcloud_firewall rule and port range
### What happened?

According to docs port range can be defined with ":"
```
port - (Required, string) Port of the Firewall Rule. Required when protocol is tcp or udp. You can use any to allow all ports for the specific protocol. Port ranges are also possible: 80:85 allows all ports between 80 and 85.
```
https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/firewall

In fact, terraform will fail:
```
      + rule {
          + destination_ips = []
          + direction       = "in"
          + port            = "1:20"
          + protocol        = "tcp"
          + source_ips      = [
              + "0.0.0.0/0",
              + "::/0",
            ]
        }

module.firewall.hcloud_firewall.remote_access: Creating...
╷
│ Error: invalid input in field 'rules' (invalid_input)
```

and port range works with "-" symbol:
```
  rule {
    direction  = "in"
    protocol   = "tcp"
    port       = "1-20"
    source_ips = ["0.0.0.0/0", "::/0"]
  }
```

hcloud_firewall resource should be created, bug in documentation.

```
  rule {
    direction  = "in"
    protocol   = "tcp"
    port       = "1-20"
    source_ips = ["0.0.0.0/0", "::/0"]
  }
```

szEvEz:
Fixes https://github.com/hetznercloud/terraform-provider-hcloud/issues/418

### DIFF
--- a/website/docs/r/firewall.html.md
+++ b/website/docs/r/firewall.html.md
@@ -16,13 +16,24 @@ Provides a Hetzner Cloud Firewall to represent a Firewall in the Hetzner Cloud.
 resource "hcloud_firewall" "myfirewall" {
   name = "my-firewall"
   rule {
-   direction = "in"
-   protocol  = "icmp"
-   source_ips = [
+    direction = "in"
+    protocol  = "icmp"
+    source_ips = [
       "0.0.0.0/0",
       "::/0"
-   ]
+    ]
   }
+
+  rule {
+    direction = "in"
+    protocol  = "tcp"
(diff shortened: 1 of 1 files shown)


LKaemmerling:
Thank you! We completely missed to change it here. Using : was our initial plan (in the development of the firewalls)

PR #426, opened 2021-08-16
Placement groups

Adi146:


website/docs/d/placementgroup.html.md
```diff
@@ -0,0 +1,34 @@
+---
```
LKaemmerling, line 1: Reference in `hcloud.erb` missing (for the other new docs too)

PR #498, opened 2022-01-15
Add missing argument for resource hcloud_ssh_key

dhoppe:
When I run `terraform apply` for the second time, an empty map is created. 
```
  ~ resource "hcloud_ssh_key" "default" {
        id          = "5356555"
      + labels      = {}
        name        = "terraform"
        # (2 unchanged attributes hidden)
    }
```

Looks like this argument exists, but has not been documented. 
```
  ~ resource "hcloud_ssh_key" "default" {
        id          = "5356555"
      ~ labels      = {
          + "environment" = "demo"
        }
        name        = "terraform"
        # (2 unchanged attributes hidden)
    }
```


fhofherr:
Hi,

good catch. Thank you for your contribution.

PR #501, opened 2022-01-26
Add resourceFloatingIPAssignmentUpdate

### LINKED ISSUE #497: [Feature]: Please allow modifying `hcloud_floating_ip_assignment.server_id` without recreation

Currently, if a Terraform configuration uses `hcloud_floating_ip_assignment`, and the IP needs to be reassigned to a different server, Terraform will recreate the `hcloud_floating_ip_assignment` resource. Implementation-wise, this seems to be due to `ForceNew` being set on the property (and there not being a handler for modifying it):

https://github.com/hetznercloud/terraform-provider-hcloud/blob/49b6cbb33629f247ab16632ab7eec4bf7523df0a/internal/floatingip/resource_assignment.go#L37

This becomes an issue when we want to use `create_before_destroy` on an `hcloud_server` resource, in order to perform a hot handover of the floating IP to a new instance of a server. Although `create_before_destroy` works as expected on the `hcloud_server` resource, the `hcloud_floating_ip_assignment` object is destroyed *immediately*, and only recreated once the new server is up.

This could be avoided by allowing the `hcloud_floating_ip_assignment` resource to be updated in-place if its `server_id` field changes. In this case, Terraform will wait until the new server is created and then update the IP assignment resource with the new server ID.

Of note is that the `hcloud_floating_ip` resource does not have this issue, but some circumstances require using `hcloud_floating_ip_assignment` to avoid circular dependencies.

CyberShadow:
- Fixes #497.

I've tested this on our HCloud project and it seems to work as described.

### DIFF
--- a/internal/floatingip/resource_assignment.go
+++ b/internal/floatingip/resource_assignment.go
@@ -21,6 +21,7 @@ func AssignmentResource() *schema.Resource {
 	return &schema.Resource{
 		CreateContext: resourceFloatingIPAssignmentCreate,
 		ReadContext:   resourceFloatingIPAssignmentRead,
+		UpdateContext: resourceFloatingIPAssignmentUpdate,
 		DeleteContext: resourceFloatingIPAssignmentDelete,
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
@@ -34,7 +35,6 @@ func AssignmentResource() *schema.Resource {
 			"server_id": {
 				Type:     schema.TypeInt,
 				Required: true,
-				ForceNew: true,
 			},
 		},
 	}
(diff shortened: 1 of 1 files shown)


LKaemmerling:
Hey @CyberShadow,

thank you for your contribution! Could you please add a test case for this [here](https://github.com/hetznercloud/terraform-provider-hcloud/blob/master/internal/e2etests/floatingip/resource_assignment_test.go)?

CyberShadow:
Thanks!

> Could you please add a test case for this [here](https://github.com/hetznercloud/terraform-provider-hcloud/blob/master/internal/e2etests/floatingip/resource_assignment_test.go)?

As I understand:

- The existing tests should cover this addition. As the test changes the server that the assignment is bound to, it should activate the code added here.

- More elaborate tests will only be meaningful in acceptance testing mode.

Regarding the latter, I'm not really sure about the process of configuring and running them. Do I need to first configure a separate Hetzner Cloud project, so that the tests don't interfere with the production VMs? Is it possible to run only a subset of tests, to avoid unnecessary expenses?

CyberShadow:
Hi @LKaemmerling,

Still waiting for your response. Please let me know how you'd like to proceed.

Thanks!

gc-ss:
> Do I need to first configure a separate Hetzner Cloud project, so that the tests don't interfere with the production VMs?

Yes

> Is it possible to run only a subset of tests, to avoid unnecessary expenses?

How much do you estimate the expenses would be?

We can request @LKaemmerling to approve a 20E credit - the typical referral credits if you refer a new "test" account from your current cloud account

LKaemmerling:
Hey @CyberShadow,

thank you for your MR! Good job!

PR #510, opened 2022-02-23
Fix error in Example Usage

carcoco:
Please consider this fix of error.


carcoco:
I think this PR is not accurate, then i close it.  Excuse me.

PR #525, opened 2022-05-06
DataSource Network `id` should be an integer

guineveresaenger:
Hi there!

As per the [schema definition](https://github.com/hetznercloud/terraform-provider-hcloud/blob/main/internal/network/data_source.go#L30), the `id` field in the documentation example should be an integer.

We've run across some doc discrepancies in our downstream Pulumi provider. Thank you so much for considering this pull request.

### DIFF
--- a/website/docs/d/network.html.md
+++ b/website/docs/d/network.html.md
@@ -11,7 +11,7 @@ This resource is useful if you want to use a non-terraform managed network.
 ## Example Usage
 ```hcl
 data "hcloud_network" "network_1" {
-  id = "1234"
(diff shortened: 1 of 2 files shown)


guineveresaenger:
Would love any insight on why this is failing CI :)

guineveresaenger:
It turns out that our docs problems were caused by incomplete pluralization in the `get_networks` data source, rather than the quotes around the example ID string.

Thank you again!

LKaemmerling:
Hey @guineveresaenger,

good catch! Thank you!

PR #560, opened 2022-08-22
feature: update workflow & golang to newest stable release 1.19

komandar:
This PR upgrades the golang version to the current available version (1.19) and the workflows action versions.


.github/workflows/test.yml
```diff
@@ -7,21 +7,21 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        terraform_version: [ "1.0.x" ]
+        terraform_version: [ "1.2.x" ]
```
LKaemmerling: I would prefer to add all three latest supported versions so:
```suggestion
        terraform_version: [ "1.0.x", "1.1.x", "1.2.x" ]

LKaemmerling (changes requested):
Hey @komandar,

thank you! I had a small suggestion. Also, the golangci-lint tests fail, can you maybe have a look there?

komandar:
> Hey @komandar,
> 
> thank you! I had a small suggestion. Also, the golangci-lint tests fail, can you maybe have a look there?

Hi @LKaemmerling 
Thanks for your feedback. I have added the other Terraform versions. Unfortunately the test cases don't seem to work with the older versions. I don't have as much insight into your CI as you do. Do you have a spontaneous idea what the problem could be?

LKaemmerling:
@komandar thank you! Good job!

PR #563, opened 2022-08-24
ci: fix not available gpg_private_key in workflow

komandar:
It seems that the import GPG key step does not get to the private key.
See: https://github.com/hetznercloud/terraform-provider-hcloud/actions/runs/2919525089

As long as the key in the repo is set correctly, it should now be able to get the data again.


LKaemmerling (approved):
Thanks!

Issue #515, opened 2022-04-06
[Docs]: <id> tag for import not visible on documentation website


der-eismann:
The documentation for import purposes uses `<id>` most of the time, e.g. [here](https://github.com/hetznercloud/terraform-provider-hcloud/blob/master/website/docs/r/server.html.md#import). However this is not visible on the [terraform documentation website](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/server#import), probably because it is interpreted as HTML.

I would change it to `123456` or a random number instead, like it is done for the [server network resource](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/server_network#import).

der-eismann:
Fixed by #564

PR #564, opened 2022-08-26
Remove < and > signs in import examples

ekeih:
The `<id>` in the import examples are not rendered correctly on the Terraform registry website, e.g. at https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/primary_ip#import

I suspect that the < and > characters are lost during some sanitization process. By removing them we should get rid of the missleading examples.

Also remove a deadlink in the readme.


ekeih:
Just noticed, that this should solve https://github.com/hetznercloud/terraform-provider-hcloud/issues/515

LKaemmerling:
Thanks Max! Good catch

PR #567, opened 2022-08-27
fix: wrong required statement

### LINKED ISSUE #543: [Bug]: documentation for hcloud_server_network: alias_ips really required?
### What happened?

I successfully applied following code:
```
resource "hcloud_server_network" "somename" {
  server_id = hcloud_server.somename.id
  network_id = hcloud_network.web.id
  ip = "10.11.1.5"
}
```

Since _alias_ips_ is required, I expected it to fail. But it applied successfully.

see in "What happened"

komandar:
In the source code the value is flagged as optional (which makes sense):
internal/server/resource_network.go
line 59

The error in the docs have been fixed accordingly.
Closes #543

### DIFF
--- a/website/docs/r/server_network.html.md
+++ b/website/docs/r/server_network.html.md
@@ -39,7 +39,7 @@ resource "hcloud_server_network" "srvnetwork" {
 ## Argument Reference
 
 - `server_id` - (Required, int) ID of the server.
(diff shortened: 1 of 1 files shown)


LKaemmerling:
Good catch! Thank you!

PR #568, opened 2022-08-27
style: unify the bool and boolean type in the docs

komandar:
Currently the type bool and boolean are used in the documentation for the same type, but due to the change only bool is used.


LKaemmerling:
Good catch! Thank you!

PR #597, opened 2022-11-23
fix(server): migration to explicit primary IPs fails

### LINKED ISSUE #577: [Bug]: importing primary IPs generally buggy
### What happened?

Since the new flexible netowrking was introduced, I wanted to import primary IPs that already existed for my server, and also reflect that change for "hcloud_server" that already had these IPs assigned. Couple of unexpected things happened:
- on one of the applies, the ipv6 resource returned an error "invalid IP address" (the apply only changed the name, auto-delete and deletion protection parameters for this IP.). Given the fact that this was simply imported and I never had to actually provide this IP, this is a bit bizarre
- subsequent apply completed succesfully for the primary IPs, but hcloud_server reported "Error: All Floating IPs of type IPv4 have to be unassigned from Server (service_error)". The server does in fact have a floating IP assigned, but considering that I'm just adding a "public_net" block to the server, containing primary IPs that it already has assigned, I didn't expect the apply to actually do anything - the server already has these primary IPs in place. (Furthermore, this apply left the server turned off, which was also undesirable).

I expected the apply after the import to succeed, and effectively only change a bunch of parameters to the primary IPs, and do nothing for the hcloud_server.

I think what would recreate my situation would be first to create this:

```
resource "hcloud_floating_ip" "main" {
  home_location = "fsn1"
  type = "ipv4"
}

resource "hcloud_floating_ip_assignment" "main_to_ip" {
  floating_ip_id = hcloud_floating_ip.main.id
  server_id      = hcloud_server.main.id
}

resource "hcloud_server" "main" {
  name        = "test"
  image       = "ubuntu-20.04"
  server_type = "cx11"
  location    = "fsn1"
}
```

Then once that is applied and succesfully created, adding this:
```
resource "hcloud_primary_ip" "main" {
  for_each = toset(["ipv4", "ipv6"])
  
  name          = "main_primary_${each.key}"
  datacenter    = "fsn1-dc14"
  type          = each.key
  assignee_type = "server"
  
  delete_protection = true
  auto_delete       = false
}
```
and this block to hcloud_server:
```
public_net {
    ipv4 = hcloud_primary_ip.main["ipv4"].id
    ipv6 = hcloud_primary_ip.main["ipv6"].id
  }
```

The import:
```
terraform import 'hcloud_primary_ip.main["ipv4"]' [ID]
terraform import 'hcloud_primary_ip.main["ipv6"]' [ID]
```
And then terraform apply causes things I'm describing (so first apply returns invalid IP, second apply complains about the floating IP and turns off the server, turning server back on manually and then doing apply works fine... except setting auto_delete doesn't work but that seems to be issue #565 and I'm assuming the fix wasn't released yet.

apricote:
When migrating a server from the previously used implicit primary IPs to using the `hcloud_primary_ip` resources and specifying those in the `hcloud_server.public_net` argument failed, because the state for the `public_net` was not correctly populated. This caused the servers to require an unnecessary re-attach of the primary ips.

Fixes #577

### DIFF
--- a/internal/server/resource.go
+++ b/internal/server/resource.go
@@ -151,6 +151,7 @@ func Resource() *schema.Resource {
 			"public_net": {
 				Type:     schema.TypeSet,
 				Optional: true,
+				Computed: true,
 				DiffSuppressFunc: func(_, _, _ string, d *schema.ResourceData) bool {
 					// Diff is only valid if "public_net" resource is set in
 					// terraform configuration.
@@ -1123,6 +1124,10 @@ func getServerAttributes(d *schema.ResourceData, s *hcloud.Server) map[string]in
 		res["network"] = networkToTerraformNetworks(s.PrivateNet)
 	}
 
+	// Wrap in slice, as type of public_net is Set, even though only one element
+	// can be set.
+	res["public_net"] = []map[string]interface{}{publicNetToTerraformPublicNet(s.PublicNet)}
+
 	if s.PlacementGroup != nil {
 		res["placement_group_id"] = s.PlacementGroup.ID
 	}
@@ -1145,6 +1150,22 @@ func networkToTerraformNetworks(privateNetworks []hcloud.ServerPrivateNet) []map
 		tfPrivateNetworks[i] = tfPrivateNetwork
 	}
 	return tfPrivateNetworks
+}
+
+func publicNetToTerraformPublicNet(publicNet hcloud.ServerPublicNet) map[string]interface{} {
+	tfPublicNet := make(map[string]interface{})
+
+	if !publicNet.IPv4.IsUnspecified() {
+		tfPublicNet["ipv4_enabled"] = true
+		tfPublicNet["ipv4"] = publicNet.IPv4.ID
+	}
+
+	if !publicNet.IPv6.IsUnspecified() {
+		tfPublicNet["ipv6_enabled"] = true
+		tfPublicNet["ipv6"] = publicNet.IPv6.ID
+	}
+
+	return tfPublicNet
 }
 
 func getPlacementGroup(ctx context.Context, c *hcloud.Client, id int) (*hcloud.PlacementGroup, error) {


apricote:
Closing as I have not managed to fix the bugs that came with my changes. This will be easier to implement after #752

apricote:
A more robust implementation of `publicNetToTerraformPublicNet` that avoids empty sets:

```go
func publicNetToTerraformPublicNet(publicNet hcloud.ServerPublicNet) map[string]interface{} {
	tfPublicNet := make(map[string]interface{})

	ipv4Enabled := !publicNet.IPv4.IsUnspecified()
	tfPublicNet["ipv4_enabled"] = ipv4Enabled
	if ipv4Enabled {
		tfPublicNet["ipv4"] = publicNet.IPv4.ID
	}

	ipv6Enabled := !publicNet.IPv6.IsUnspecified()
	tfPublicNet["ipv6_enabled"] = ipv6Enabled
	if ipv6Enabled {
		tfPublicNet["ipv6"] = publicNet.IPv6.ID
	}

	return tfPublicNet
}
```

apricote:
Adding `computed: true` to the attribute causes issues with the transition from an explicitly defined `public_net` to a removed `public_net`.

- when the user removes the `public_net` block, the default should be restored. The default is generated random IPv4 and IPv6
- if we mark the field as `computed`, Terraform will not mark the fields as changed and keep the values we previously wrote to the state.
- Because of this, we can not detect that the user has removed the attribute, and can not trigger the logic to generate random IPs

This can only be solved by the plugin framework, which gives us access to the precise configuration that the user specified: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource#UpdateRequest

jooola:
Reopening as the issue is still relevant, we are just being blocked by the migration to the plugin framework.

jooola:
Nevermind, I thought it was an issue, not a PR.

PR #606, opened 2022-12-25
docs: Add missing location (hil)

akirak:
Hillsboro, OR (us-west) is currently missing from the documentation, so I have added it.


apricote:
Thanks for the contribution @akirak!

PR #638, opened 2023-03-13
chore: Test against Terraform v1.4

apricote:
The version was released last week.

Additionally I removed all tests for older versions of terraform, as terraform promises compatibility for the interaction with providers.


apricote:
I expect this change to help with the severe rate limiting I am seeing on the other currently open PRs.

PR #660, opened 2023-04-13
ci: run e2etests in parallel

apricote:
Most of the time in e2etests is spent waiting for async actions in the Hetzner Cloud API. By running the tests in parallel we should be able to shave of most of that time.


apricote:
This roughly halves the time spent on integration tests from ~20 minutes before down to ~10 minutes.

samcday (approved):

PR #691, opened 2023-06-12
feat: expose deprecation info for server types

apricote:
- Add new fields to `data.hcloud_server_type` & `data.hcloud_server_types` that include the deprecation info
- Log warning when the user creates a server with a deprecated server type


jonasbadstuebner (approved):
:+1:

Issue #751, opened 2023-09-14
[Feature]: (Option to) Shutdown server gracefully when destroyed before deleting it


LionC:
### What whould you like to see?

Currently, when destroying an `hcloud_server` resource, the server is outright deleted via the API ([happening here I think](https://github.com/hetznercloud/terraform-provider-hcloud/blob/f72cd3bdac6acc1b80bc20ffdf766f52a3385d2d/internal/server/resource.go#L852)) without shutting it down properly.

This causes services that might want to terminate gracefully to not have the chance to do so. It would be nice if it was possible to shut down the server properly before deleting it - maybe via an option in the resource if that change would be considered breaking otherwise.

I would be happy to attempt an implementation if this is something the team would approve.

apricote:
In general this sounds like a nice option to have! One issue I can think of (see https://github.com/hetznercloud/cli/issues/489) is that "graceful shutdown" is not implemented directly in the API. There are two endpoints which do this:

-  [power off](https://docs.hetzner.cloud/#server-actions-power-off-a-server): once issued, this will immediately stop the server without giving it any time to process the shut down. AFAIK this is what happens if the server is deleted.
- [shut down](https://docs.hetzner.cloud/#server-actions-shutdown-a-server): this will send the ACPI shutdown request to the server. The action that is returned from this endpoint only means "the request was sent", not that the server actually shut down. The OS is responsible for handling the ACPI request and we do not really know when it was processed. Only solution I can think of is a timeout, after which we proceed with the deletion, or polling the server object to watch for `status: off`

LionC:
> * [shut down](https://docs.hetzner.cloud/#server-actions-shutdown-a-server): this will send the ACPI shutdown request to the server. The action that is returned from this endpoint only means "the request was sent", not that the server actually shut down. The OS is responsible for handling the ACPI request and we do not really know when it was processed. Only solution I can think of is a timeout, after which we proceed with the deletion, or polling the server object to watch for `status: off`

Polling would probably be the more desirable solution, maybe with a configurable timeout with a sensible default. But a hard timeout after which the server is deleted would also be better than no solution.

I would be happy to try an implementation for the polling solution.

apricote:
> I would be happy to try an implementation for the polling solution.

That would be awesome! Feel free to open a draft PR if you have any questions :)

LionC:
[...]

PR #755, opened 2023-09-16
feat(server): Optional graceful shutdown before deleting servers

LionC:
Solves #751

### DIFF
--- a/internal/server/resource.go
+++ b/internal/server/resource.go
@@ -888,7 +888,7 @@ func resourceServerDelete(ctx context.Context, d *schema.ResourceData, m interfa
 			// If shutting down does not work, add a warning and move on with deletion
 			warnings = append(warnings, diag.Diagnostic{
 				Severity: diag.Warning,
-				Summary:  fmt.Sprintf("Server id %d took longer than 30s to shut down gracefully, deleting it anyways.", serverID),
+				Summary:  fmt.Sprintf("Server id %d did not finish shutting down gracefully in time, deleting it anyways.", serverID),
 			})
 
 			err = nil
--- a/internal/server/resource.go
+++ b/internal/server/resource.go
@@ -868,11 +868,11 @@ func resourceServerDelete(ctx context.Context, d *schema.ResourceData, m interfa
 			return hcclient.ErrorToDiag(err)
 		}
 
-		// We give the server some time to shut down
+		// Give the server some time to shut down
 		err = control.Retry(control.DefaultRetries, func() error {
 			server, _, err := client.Server.GetByID(ctx, serverID)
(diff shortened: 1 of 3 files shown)


internal/server/resource.go
```diff
@@ -849,6 +854,47 @@ func resourceServerDelete(ctx context.Context, d *schema.ResourceData, m interfa
...
+
+			// Server has shut down successfully
+			return nil
+		})
+		if err != nil {
+			// If shutting down does not work, add a warning and move on with deletion
+			warnings = append(warnings, diag.Diagnostic{
+				Severity: diag.Warning,
+				Summary:  fmt.Sprintf("Server id %d took longer than 30s to shut down gracefully, deleting it anyways.", serverID),
+			})
```
apricote: Nice!

website/docs/d/server.html.md
```diff
@@ -47,3 +47,4 @@ data "hcloud_server" "s_3" {
 - `placement_group_id` - (Optional, string) Placement Group ID the server is assigned to.
 - `delete_protection` - (bool) Whether delete protection is enabled.
 - `rebuild_protection` - (bool) Whether rebuild protection is enabled.
+- `shutdown_before_deletion` - (bool) Whether to try shutting the server down gracefully before deleting it.
```
apricote: You added this argument to the datasource docs. It should go into the `Argument` section of the resource docs at `website/docs/r/server.html.md`

apricote (changes requested):
One issue with the docs, otherwise this looks great :)

internal/server/resource.go
LionC: I actually re-read the text because of that comment and changed it, removing the time, as I am using the retry function's defaults instead of a set timer.

website/docs/d/server.html.md
LionC: Good catch, should be fixed now (added to both arguments and attributes)

apricote (approved):

LionC:
@apricote thanks for the approval. The failing check looks like a CI setup problem to me.

Can you give me an idea on when this will be available in a release? I want to use the feature in our project :-)

apricote:
Yea, E2E CI is expected to fail right now for external contributors. I tested your previous changes locally, so I think this should be good to go. E2E tests will then run in `main`.

apricote:
> Can you give me an idea on when this will be available in a release? I want to use the feature in our project :-)

I can publish a new release once the `main` pipeline succeeded.

apricote:
Looks like I was wrong and it actually broke the e2e tests. Pushed a fix at #756.

PR #797, opened 2023-11-21
test: replace null_resource with terraform_data

jooola:
As suggested in https://github.com/hashicorp/terraform-plugin-framework/issues/875#issuecomment-1820916537

### DIFF
--- a/internal/testdata/r/terraform_data_resource.tf.tmpl
+++ b/internal/testdata/r/terraform_data_resource.tf.tmpl
@@ -2,7 +2,7 @@
 // changes in resources, not in datasources or outputs.
 // You can pass any [testtemplate.DataCommon] to it.
 
-resource "null_resource" "{{ .RName }}" {
+resource "terraform_data" "{{ .RName }}" {
   triggers = {
     resource = jsonencode({{ .TFID }})
   }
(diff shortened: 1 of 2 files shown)


internal/testdata/r/terraform_data_resource.tf.tmpl
```diff
@@ -2,7 +2,7 @@
 // changes in resources, not in datasources or outputs.
 // You can pass any [testtemplate.DataCommon] to it.
 
-resource "null_resource" "{{ .RName }}" {
+resource "terraform_data" "{{ .RName }}" {
   triggers = {
```
apricote, line 6:
```suggestion
  input = {
```

See https://developer.hashicorp.com/terraform/language/resources/terraform-data#input

This fails the e2e tests currently.

apricote (changes requested):

internal/testdata/r/terraform_data_resource.tf.tmpl
```diff
@@ -2,7 +2,7 @@
...
     resource = jsonencode({{ .TFID }})
   }
```
apricote:
```suggestion
  input = {{ .TFID }}
```

This should work according to the new docs for testing datasources: https://github.com/hashicorp/terraform-plugin-framework/pull/877/files#diff-577539d2f18c0175da7dcefd0cc00d19c28cfbce5f444b08a3a595c018503d25R108

apricote (changes requested):

apricote (approved):

Issue #752, opened 2023-09-15
[Feature]: Migrate to TF Plugin Framework


apricote:
### What whould you like to see?

Terraform has launched a new "Plugin Framework" that replaces the Plugin SDK v2. This new framework supports some advanced scenarios that currently do not work properly. It can help us solve a number of bugs that we are experiencing:

- SDK v2 only supports "int" while Plugin Framework supports explicit "int64" types (Related to [hcloud-go v2](https://github.com/hetznercloud/hcloud-go/releases/tag/v2.0.0), #997, #877)
- Plugin Framework supports strongly typed access to fields, which will help with a number of issues I discovered while trying to migrate to hcloud-go v2
- Terraform 1.4 introduced a bug for nested sets with computed fields (https://github.com/hetznercloud/terraform-provider-hcloud/issues/650). This would be fixed by switching to Plugin Framework
- Plugin Framework has a number of util functions regarding Diagnostics (https://github.com/hetznercloud/terraform-provider-hcloud/issues/692) that make them easier to use.
- SDK v2 only supports binary `ForceNew` for schema attributes. For `hcloud_server.server_type` it makes sense to `ForceNew` when switching between x86 and Arm, but to keep the same server and scale it in case the architecture stays the same. This is supported by the Plugin Framework. See #853
- For `hcloud_server.public_net` we have some bugs that are caused by the "Set" type for this argument. These can only be fixed in the Plugin Framework. See https://github.com/hetznercloud/terraform-provider-hcloud/issues/944 and https://github.com/hetznercloud/terraform-provider-hcloud/pull/597#issuecomment-2328321946

This migration will be a huge refactoring effort. We plan on doing it in many smaller steps, migrating a couple of resources at a time.

- [x] #749
- [ ] Migrate individual resources/datasources
  - [ ] Certificates
    - [ ] `hcloud_certificate`
    - [ ] `hcloud_uploaded_certificate`
    - [ ] `hcloud_managed_certificate`
    - [ ] `data.hcloud_certificate`
    - [ ] `data.hcloud_certificates`
  - [x] Datacenter
    - [x] `data.hcloud_datacenter` #779
    - [x] `data.hcloud_datacenters` #779
  - [ ] Firewall
    - [ ] `hcloud_firewall`
    - [ ] `hcloud_firewall_attachment`
    - [ ] `data.hcloud_firewall`
    - [ ] `data.hcloud_firewalls`
  - [ ] Floating IPs
[...]

PR #798, opened 2023-11-22
feat(provider): choose between constant & exponential backoff for actions

beagleknight:
This changes the poll backoff function to use `ConstantBackoff` instead of `ExponentialBackoff`. The current `poll_interval` option is a bit misleading because assumes a constant polling interval instead of an exponential one. This causes to wait much more time in some scenarios where you requested the resource status to the API a few times.

In scenarios where you expect a specific amount of time, such as creating a cloud instance with a custom snapshot, this change makes the behavior more predictable.

### DIFF
--- a/hcloud/plugin_provider.go
+++ b/hcloud/plugin_provider.go
@@ -7,11 +7,13 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/provider"
 	"github.com/hashicorp/terraform-plugin-framework/provider/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 
@@ -56,6 +58,13 @@ func (p *PluginProvider) Schema(_ context.Context, _ provider.SchemaRequest, res
 				Description: "The interval at which actions are polled by the client. Default `500ms`. Increase this interval if you run into rate limiting errors.",
 				Optional:    true,
 			},
+			"poll_function": schema.StringAttribute{
+				Description: "The type of function to be used during the polling.",
+				Optional:    true,
+				Validators: []validator.String{
+					stringvalidator.OneOf([]string{"constant", "exponential"}...),
+				},
+			},
 		},
 		// TODO: Uncomment once we get rid of the SDK v2 Provider
 		// MarkdownDescription: `The Hetzner Cloud (hcloud) provider is used to interact with the resources supported by
@@ -69,6 +78,7 @@ type PluginProviderModel struct {
 	Token        types.String `tfsdk:"token"`
 	Endpoint     types.String `tfsdk:"endpoint"`
 	PollInterval types.String `tfsdk:"poll_interval"`
+	PollFunction types.String `tfsdk:"poll_function"`
 }
 
 // Configure is called at the beginning of the provider lifecycle, when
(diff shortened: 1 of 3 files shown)


jooola:
I agree that `poll_interval` suggests a constant backoff. But we need to be cautious with such change as it might introduce rate limits errors for some users.

The origin of this change is https://github.com/hetznercloud/terraform-provider-hcloud/commit/c8d890521f44a94a6617b379922649bce35ca26b

Maybe we want to provide an extra setting such as `poll_function = oneOf('constant', 'exponential')`.


The same change must be made to the plugin provider: https://github.com/hetznercloud/terraform-provider-hcloud/blob/4811531c95f42ab42cc074a97fddfa99154d77c5/hcloud/plugin_provider.go#L124

apricote:
I agree with what @jooola said.

To explain the reasoning for switching it to exponential: The default rate limit for actions is 2 requests per second. If Terraform creates and waits for 5 resources (with refreshes every `500ms`), you can get into a situation where the rate limit is exhausted and you can not really make progress, because every request you make is just using up the limit once again.

By using an exponential backoff you avoid this issue, because you just wait for longer between the API calls.

> In scenarios where you expect a specific amount of time, such as creating a cloud instance with a custom snapshot, this change makes the behavior more predictable.

In my experience this time is highly variable, depending on the load on the snapshot storage. How long did you wait and how long did you expect it to take? Perhaps we need some kind of upper limit in the exponential backoff (maybe 30-60 seconds), to avoid situations like this.

Making constant vs. exponential configurable sounds good to me, I would prefer the exponential default though.

fcsonline:
I have to say that I share the same thoughts as @beagleknight about the expected constant behaviour, but I also understand the risk of introducing a new behaviour combined with the `500ms` that will cause much pain. The best solution, as you said is a new configuration option with the exponential default for backward compatibility.

beagleknight:
Thanks for your replies @jooola and @apricote .
I just updated the code to allow users to configure the polling function. I'm not sure if the code is ok, since I don't have much knowledge about `go`. Can you please review it?

apricote:
Thanks! I will take another look tomorrow :)

hcloud/plugin_provider.go
```diff
@@ -56,6 +56,10 @@ func (p *PluginProvider) Schema(_ context.Context, _ provider.SchemaRequest, res
 				Description: "The interval at which actions are polled by the client. Default `500ms`. Increase this interval if you run into rate limiting errors.",
 				Optional:    true,
 			},
+			"poll_function": schema.StringAttribute{
+				Description: "The type of function to be used during the polling.",
+				Optional:    true,
+			},
```
apricote: Could you add a `stringvalidator.OneOf` to the attribute to make sure that one of the expected values is chosen? Similar to how you did it for the other provider, but this time in the :sparkles: new and fancy way :sparkles: (see #752 for background why there is two providers).

- https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators@v0.12.0/stringvalidator#OneOf

apricote (changes requested):
This is already looking pretty good :) I have updated the PR title to match the new content (as we use it for generating the Changelog).

There is one remaining inconsistency between the two providers. Do not worry about the failed integration tests, they do not work for external contributors. If you want, you can fix the formatting (see unit tests & lint job), otherwise, I can also take care of it before merging :)

**EDIT**: [Docs](https://github.com/hetznercloud/terraform-provider-hcloud/blob/main/website/docs/index.html.markdown) are also missing, but I can also do them before merging if you dont want to.

beagleknight:
Hey @apricote ! Thanks for your comments. I already addressed them and squashed all commits. Can you review it when you have some time? Thank you!

hcloud/provider.go
```diff
@@ -68,6 +68,13 @@ func Provider() *schema.Provider {
 				Default:     "500ms",
 				Description: "The interval at which actions are polled by the client. Default `500ms`. Increase this interval if you run into rate limiting errors.",
 			},
+			"poll_function": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "exponential",
+				Description:  "The type of function to be used during the polling.",
+				ExactlyOneOf: []string{"constant", "exponential"},
```
apricote:
```suggestion
				ValidateFunc: validation.StringInSlice([]string{"constant", "exponential"}, false),
```

Sorry for this late find, `ExactlyOneOf` is used to tell terraform that one of the specified fields must be set, not that this field needs to be one of two values.

apricote (changes requested):
One more issue i found through the unit tests.

I tried to fix and push to the branch myself, but looks like this is disabled.

hcloud/provider.go
beagleknight: Got it! Thanks @apricote

apricote (changes requested):
Tests are still failing because an import for `"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"` is missing in `hcloud/provider.go`, could you add this and run `make test` locally to verify that it works?

beagleknight:
@apricote Sorry for the late response! I just added the missing import to the code (and finally installed the devtools in vscode 😄 )

jooola (approved):

apricote (approved):
Thank you very much @beagleknight! Will be included in the next release :)

beagleknight:
Hey @apricote ! Do you know when the next version will be released? We would like to use this change in our template :). Thanks!

apricote:
> Hey @apricote ! Do you know when the next version will be released? We would like to use this change in our template :). Thanks!

Released in [v1.45.0](https://github.com/hetznercloud/terraform-provider-hcloud/releases/tag/v1.45.0)

fcsonline:
Thank you so much for the release, @apricote !! Testing it right now.

Issue #683, opened 2023-05-22
[Feature]: return disksize from hcloud_server resource/datasource


harmesit:
### What whould you like to see?

currently, you are able to not scale the Disk when scaling the rest of the Server resources which is nice for later downgrade. doing so in the Webui does not change the data returned for the resource or datasource.

this leads to inconsistencies when using terraform to automatically document the vm aside from just creating and managing it when documenting the available resources via the server_type datasource only.

EttoreFoti:
I would like to work on it.

edit: https://github.com/hetznercloud/terraform-provider-hcloud/pull/801

This should be enough to add the data you're interested in.

apricote:
Implemented in #801

PR #801, opened 2023-11-27
feat(server): add `primary_disk_size` attribute in resource and datasource

EttoreFoti:
Resolve enhancement request in issue #683.

### DIFF
--- a/internal/server/data_source.go
+++ b/internal/server/data_source.go
@@ -105,6 +105,10 @@ func getCommonDataSchema() map[string]*schema.Schema {
 			Type:     schema.TypeBool,
 			Computed: true,
 		},
+		"primary_disk_size": {
+			Type:     schema.TypeInt,
+			Computed: true,
+		},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


EttoreFoti:
Integrations test failing because it looks like a repo secret is missing.

apricote (changes requested):
Very nice!

Could you add the new attribute to the docs pages?

- https://github.com/hetznercloud/terraform-provider-hcloud/blob/fa7ea1f167750e64cc9ebbb26ce0eb8171d55df6/website/docs/r/server.html.md?plain=1#L192
- https://github.com/hetznercloud/terraform-provider-hcloud/blob/fa7ea1f167750e64cc9ebbb26ce0eb8171d55df6/website/docs/d/server.html.md?plain=1#L31

apricote:
Integration tests are expected to fail. GitHub makes it very hard to allow access to Secrets to external contributors. We are working on an alternative solution :)

EttoreFoti:
Changes added, thanks for the explanation :)

apricote (approved):
I pushed another commit to move the attribute to the right section, it was previously in the explanation of the `network` block.

Thanks a lot for the new feature @EttoreFoti!

Issue #715, opened 2023-07-11
[Feature]: Ability to directly add server IPs to firewall rules without having to do string interpolation


mzhaase:
### What whould you like to see?

The following cases are all very common and should work without having to do string interpolation:

```HCL
resource "hcloud_firewall" "this" {
    [...]
    # this currently gives an error because it is not a CIDR, requiring some
    # string interpolation magic to work
    rule {
        direction  = "in"
        protocol   = "tcp"
        port       = "443"
        source_ips = hcloud_server.this[*].ip
   }

    # this currently gives an error because network is a set. But adding private
    # IP addresses to a firwall rule is a very common use case that should be
    # supported. Right now we have to use a local to create the CIDRs but
    # this way we lose the ordering between the firewall rule and the server
    rule {
        direction  = "in"
        protocol   = "tcp"
        port       = "443"
        source_ips = hcloud_server.this[*].network[0].ip
    }
```

Of course even better would be to be able to use source_firewall_ids like on AWS
but that is probably a different request.

PR #807, opened 2023-12-04
feature(firewall): add ability to directly add server IPs to firewall rules

EttoreFoti:
Resolve improvement requested in #715


internal/firewall/resource.go
```diff
@@ -117,7 +117,15 @@ func Resource() *schema.Resource {
...
-									return strings.ToLower(i.(string))
+									ipS := strings.ToLower(i.(string))
+									if !strings.Contains(ipS, "/") {
+										if strings.Contains(ipS, ":") {
+											ipS += "/64"
+										} else if strings.Contains(ipS, ".") {
+											ipS += "/32"
+										}
+									}
+									return ipS
```
apricote, line 128: Using string manipulation for this feels wrong and brittle to me.

Can we instead use the [`net` package](https://pkg.go.dev/net) and its methods and structs `ParseCIDR`, `ParseIP`, `IPNet` to implement this?

Some quick pseudo code:

```go
_, ipnet, err := net.ParseCIDR(i.(string))
if err != nil {
  ip := net.ParseIP(i.(string))
  if ip == nil {
    // No CIDR or IP
    return
  }
  if ip.To4() != nil {
    // IPv4
    ipnet = &net.IPNet{
      IP: ip,
      IPMask: net.CIDRMask(32, 32)
    }
  } else {
     // If To4 returns nil, IP is not IPv4 => IPv6
    ipnet = &net.IPNet{
      IP: ip,
      IPMask: net.CIDRMask(64, 128)
    }
  }

  return ipnet.String()
}
```

apricote (changes requested):
Thank you very much for the PR @EttoreFoti!

In addition to the one requested change, could you move the logic to a function, to avoid duplicating it three times?

apricote:
Hey @EttoreFoti,

a fix for another bug was closely related to this PR. I have implemented both together in #874 and added you as a co-author to the commit.

PR #835, opened 2024-01-09
ci: split test logs per test

jooola:
Generates a new logs file for each test we run.

Leverage the TF_LOG_PATH_MASK env variable, which format the string with a test name slug, e.g.: `fmt.Sprintf("test-%s.log", testNameSlug)`

https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing@v1.6.0/helper/logging#pkg-constants


jooola:
image

PR #848, opened 2024-01-22
ci: fix coverage of e2e tests not reported

apricote:
By default "go test" only considers the coverage of code in the same package as the test. As our e2e tests are located in different packages to the actual resources, this always resulted in 0% coverage.

I also added codecov flags to e2e & unit tests so we can differentiate between the coverage in codecov UI.


jooola (approved):

apricote:
Coverage works :100: it now also shows the coverage for unit vs e2e tests in the codecov comment.

PR #855, opened 2024-01-29
refactor: add resource labels schema attribute utils

jooola:
Add shared labels schema attribute for the plugin framework migration.

### DIFF
--- a/internal/util/resourceutil/labels.go
+++ b/internal/util/resourceutil/labels.go
@@ -1,0 +1,48 @@
+package resourceutil
+
+import (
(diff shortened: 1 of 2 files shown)


apricote (approved):
E2E test failures are known flakes.

jooola:
Hmm, we might have introduced more flakkyness in this PR, the diagnostics are not always sorted so the comparaison might fail.

https://github.com/hetznercloud/terraform-provider-hcloud/actions/runs/7695348612/job/20968061286#step:8:4965

PR #859, opened 2024-01-31
test: sort diagnostics before diff

jooola:
Fix flakiness introduced in #855


jooola:
CI is failing be we need #858

PR #872, opened 2024-02-16
docs(lb): various fixes

apricote:
docs: load_balancer target block is deprecated
  The `target` block in `hcloud_load_balancer` was deprecated [3 years ago](https://github.com/hetznercloud/terraform-provider-hcloud/commit/14c100065d06717e4e3209dd64c46626dd84e917) and not even documented on the docs page. For clarity the example should not use it.

docs: fix broken load_balancer_network example
  The example does not specify the dependency on the subnet resource. This can cause a race condition when the Load Balancer is attached to the network before the subnet is created.

### DIFF
--- a/website/docs/r/load_balancer.html.md
+++ b/website/docs/r/load_balancer.html.md
@@ -8,12 +8,12 @@ description: |-
 
 # hcloud_load_balancer
 
-  Provides a Hetzner Cloud Load Balancer to represent a Load Balancer in the Hetzner Cloud.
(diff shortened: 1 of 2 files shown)


jooola (approved):

jooola (commented):
I assume you prefer not to squash this PR?

apricote:
> I assume you prefer not to squash this PR?

Nope, release-please really does not liked non-squashed PRs. I have structured the commit message so release-please can properly parse the commits from it: https://github.com/googleapis/release-please?tab=readme-ov-file#what-if-my-pr-contains-multiple-fixes-or-features

PR #874, opened 2024-02-16
feat(firewall): set IPs without prefix directly

### LINKED ISSUE #715: [Feature]: Ability to directly add server IPs to firewall rules without having to do string interpolation

### LINKED ISSUE #870: [Bug]:  hcloud_firewall is updated in place even though there are no changes
### What happened?

On every `terraform plan` or `terraform apply` terraform wants to update the firewall, even though there are no changes.

I have multiple firewall rules, but this is the only one where it keeps happening. 
The major difference between this rule and other rules is, that this rule consists of a set of two rules, to allow port 80 and 443.
I tried recreating it manually and import it, also deleting it and creating it by terraform, same behaviour.

```
  # hcloud_firewall.http will be updated in-place
  ~ resource "hcloud_firewall" "http" {
        id     = "1245890"
        name   = "http"
        # (1 unchanged attribute hidden)

      - rule {
          - description     = "Allow HTTP traffic" -> null
          - destination_ips = [] -> null
          - direction       = "in" -> null
          - port            = "80" -> null
          - protocol        = "tcp" -> null
          - source_ips      = [
              - "0.0.0.0/0",
              - "::/0",
            ] -> null
        }
      + rule {
          + description     = "Allow HTTP traffic"
          + destination_ips = []
          + direction       = "in"
          + port            = "80"
          + protocol        = "tcp"
          + source_ips      = [
              + "0.0.0.0/0",
              + "::0/0",
            ]
        }

        # (2 unchanged blocks hidden)
    }
```

I expected terraform not modifying the resource.

```terraform
terraform {
  required_providers {
    hcloud = {
      source  = "hetznercloud/hcloud"
      version = ">=1.45.0"
    }
  }
}

provider "hcloud" {
  token = var.hcloud_token
}

variable "hcloud_token" {
  description = "Hetzner Cloud API Token"
  type        = string
  sensitive   = true
}

resource "hcloud_firewall" "http" {
  name = "http"

  rule {
    description = "Allow HTTP traffic"
    direction   = "in"
    protocol    = "tcp"
    port        = 80
    destination_ips = []
    source_ips = [
      "0.0.0.0/0",
      "::0/0"
    ]
  }

  rule {
    description = "Allow HTTPS traffic"
    direction   = "in"
    protocol    = "tcp"
    port        = 443
    destination_ips = []
    source_ips = [
      "0.0.0.0/0",
      "::/0"
    ]
  }

  apply_to {
    label_selector = "http"
  }

  labels = {
    "http" = ""
  }
}

```

apricote:
This commit allows users to specify allowed IPs directly, without having to awkwardly put an `/32` at the end.

If an IP is passed, we only allow this single IP (`/32` for IPv4, `/128` for IPv6). This matches the behavior in Cloud Console.

Closes #807
Closes #715

## fix(firewall): unnecessary diff if user specified non-minimal IPv6
  If the user specified an IPv6 which was not minimal (eg. `::0` instead of `::`), the API would modify this to the minimal form. In the next run terraform reports a diff and tries to update this over and over again. We now do this normalization locally to avoid showing these diffs to users.

Closes #870

### DIFF
--- a/internal/firewall/validation_test.go
+++ b/internal/firewall/validation_test.go
@@ -30,16 +30,6 @@ func TestValidateIPDiag(t *testing.T) {
 			err:  diag.Diagnostics{diag.Diagnostic{Severity: 0, Summary: "invalid CIDR address: test", Detail: "", AttributePath: cty.Path(nil)}},
 		},
 		{
-			name: "Missing CIDR notation (IPv4)",
-			ip:   "10.0.0.0",
-			err:  diag.Diagnostics{diag.Diagnostic{Severity: 0, Summary: "invalid CIDR address: 10.0.0.0", Detail: "", AttributePath: cty.Path(nil)}},
-		},
-		{
-			name: "Missing CIDR notation (IPv6)",
-			ip:   "fe80::",
-			err:  diag.Diagnostics{diag.Diagnostic{Severity: 0, Summary: "invalid CIDR address: fe80::", Detail: "", AttributePath: cty.Path(nil)}},
-		},
-		{
 			name: "Host bit set (IPv4)",
 			ip:   "10.0.0.5/8",
 			err:  diag.Diagnostics{diag.Diagnostic{Severity: 0, Summary: "10.0.0.5/8 is not the start of the cidr block 10.0.0.0/8", Detail: "", AttributePath: cty.Path(nil)}},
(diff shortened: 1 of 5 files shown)


internal/firewall/validation_test.go
```diff
@@ -58,8 +48,58 @@ func TestValidateIPDiag(t *testing.T) {
...
 
 			if test.err != nil {
-				assert.Equal(t, err, test.err)
+				assert.Equal(t, test.err, err)
 			}
 		})
 	}
 }
+
+func Test_normalizeIP(t *testing.T) {
```
jooola, line 57: I believe we are missing the badly formatted IP address test case, should we add it ? I see this is handled later in the validateIPDiag function, but maybe a test case could make sure we test the invalid path.

jooola (approved):
Apart from the missing test case this is good to go!

internal/firewall/validation_test.go
apricote: Good catch, added a testcase for this.

PR #902, opened 2024-03-28
fix(server): cannot update placement group

jooola:
The `placement_group` field name must be `placement_group_id`, the placement group could not be updated while using the wrong field name.

When no placement group is set, we find an id of `0`, which *seem* to be expected by Terraform.


jooola:
https://github.com/hetznercloud/terraform-provider-hcloud/actions/runs/8466554523/job/23195567590?pr=902#step:8:4846

apricote (approved):

PR #903, opened 2024-03-28
test: allow saving env variables in a .env file

jooola:
Add a helper for the development of the provider.


README.md
```diff
@@ -71,3 +71,17 @@ _Note:_ Acceptance tests create real resources, and often cost money to run.
 ```
 $ make testacc
 ```
+
+You may save your acceptance tests environment variables in the `.env` file, for example:
+
+```sh
+$ cat .env
+HCLOUD_TOKEN=YOUR_API_TEST_TOKEN
+TF_ACC=1
```
apricote: If you run `test ./...` with this change, it will _always_ run the acceptance tests, not only the unit tests. For our kubernetes projects we use build constraints to solve this, but not sure how well this integrates with Goland.

jooola: I can remove the TC_ACC line. This is a free form file, people can store what they need in it.

apricote (approved):

PR #922, opened 2024-05-02
fix(ssh-key): data inconsistency with empty label objects

### LINKED ISSUE #921: [Bug]:  [hcloud_ssh_key] Error: Provider produced inconsistent result after apply
### What happened?

# opentofu-1.6.2
# terraform-1.8.2
# provider[registry.terraform.io/hetznercloud/hcloud] = 1.46.1
# provider[registry.opentofu.org/hetznercloud/hcloud] = 1.46.1

```
~$ tofu / terraform apply
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to module.ssh_key["key01"].hcloud_ssh_key.this, provider "provider[\"registry.terraform.io/hetznercloud/hcloud\"]" produced an unexpected new value: .labels: was cty.MapValEmpty(cty.String), but now
│ null.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to module.ssh_key["key02"].hcloud_ssh_key.this, provider "provider[\"registry.terraform.io/hetznercloud/hcloud\"]" produced an unexpected new value: .labels: was cty.MapValEmpty(cty.String), but now null.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

simply ssh key add as resource via module

# module "hcloud-ssh-key":

```
# -- main

resource "hcloud_ssh_key" "this" {

  lifecycle {
    ignore_changes        = [fingerprint] # TODO: bug
    create_before_destroy = false
  }

  name       = var.key_name
  public_key = var.key
  labels     = try(var.labels, null)
}

# -- vars

variable "key_name" {
  description = "Hetzner Cloud SSH key name"
  type        = string
  default     = ""
}

variable "key" {
  description = "Hetzner Cloud SSH key content"
  type        = string
  default     = ""
  sensitive   = true
}

variable "labels" {
  description = "Hetzner Cloud ssh key user-defined labels"
  type        = map(any)
  default     = {}
}
```

# module call from manifest:

```
module "ssh_key" {

  for_each = var.ssh_key

  source = "../terraform-modules/hcloud-ssh-key"

  key_name = each.value.name
  key      = each.value.key
}

variable "ssh_key" {
  description = "Hetzner Cloud SSH keys"
  type        = map(any)
  default = {
    "key01" = {
      name = "key01"
      key  = "ssh-ed25519 <key01>"
    },
    "key02" = {
      name = "key02"
      key  = "ssh-ed25519 <key02>"
    }
  }
}
```

apricote:
When we implemented the Label helper we added some extra logic that handled the difference between the `label` attribute default (null) and the Hetzner Cloud API default (`{}`). This worked well in our test cases, but breaks if you pass an empty object to the attribute:

    When applying changes to hcloud_ssh_key.this, provider "provider[\"registry.terraform.io/hetznercloud/hcloud\"]" produced an unexpected new value: .labels: was cty.MapValEmpty(cty.String), but now null.

We have now fixed this by setting the default of the labels field to an empty object to match the return value of the API. With this, we no longer need the workaround to handle null labels in Terraform config.

Fixes #921

### DIFF
--- a/internal/loadbalancer/resource_target_test.go
+++ b/internal/loadbalancer/resource_target_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
+
 	"github.com/hetznercloud/hcloud-go/hcloud"
 	"github.com/hetznercloud/terraform-provider-hcloud/internal/loadbalancer"
 	"github.com/hetznercloud/terraform-provider-hcloud/internal/network"
--- a/internal/sshkey/resource_test.go
+++ b/internal/sshkey/resource_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/hetznercloud/terraform-provider-hcloud/internal/teste2e"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+
 	"github.com/hetznercloud/hcloud-go/hcloud"
 	"github.com/hetznercloud/terraform-provider-hcloud/internal/testsupport"
 	"github.com/hetznercloud/terraform-provider-hcloud/internal/testtemplate"
--- a/internal/util/resourceutil/labels.go
+++ b/internal/util/resourceutil/labels.go
@@ -3,8 +3,10 @@ package resourceutil
 import (
 	"context"
 
+	"github.com/hashicorp/terraform-plugin-framework/attr"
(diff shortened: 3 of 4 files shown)


apricote:
We have manually tested the various combinations for empty `label` attributes:

- No label attribute
- `null`
- `{}`

We plan to add some more detailed tests in the coming weeks, but wanted to get this fix out immediately.

jooola (approved):

PR #923, opened 2024-05-02
fix(subnet): handle new error message for deleting subnets with attached resources

apricote:
The error message we rely on was never part of the API contract. It is soon being changed to the new message I have added. This does not fix any immediate bug, but prepares the provider for the coming months.

### DIFF
--- a/internal/network/resource_subnet.go
+++ b/internal/network/resource_subnet.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
 	"github.com/hetznercloud/hcloud-go/hcloud"
 	"github.com/hetznercloud/terraform-provider-hcloud/internal/control"
 	"github.com/hetznercloud/terraform-provider-hcloud/internal/util/hcloudutil"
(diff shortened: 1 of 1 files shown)


internal/network/resource_subnet.go
```diff
@@ -168,7 +169,8 @@ func resourceNetworkSubnetDelete(ctx context.Context, d *schema.ResourceData, m
 		if hcloud.IsError(err, hcloud.ErrorCodeConflict) || hcloud.IsError(err, hcloud.ErrorCodeLocked) {
 			return err
 		}
-		if hcloud.IsError(err, hcloud.ErrorCodeServiceError) && strings.Contains(err.Error(), "servers are attached") {
+		if hcloud.IsError(err, hcloud.ErrorCodeServiceError) &&
+			(strings.Contains(err.Error(), "servers are attached") || strings.Contains(err.Error(), "network has attached resources")) {
```
jooola, line 173: This feels dirty, any plan to handle this with a not human/changeable error message.

jooola (approved):

internal/network/resource_subnet.go
apricote: I do agree, relying on the message is hacky and can (evidently) break from time to time. I do not see any nice workarounds for this though at this time.

Changing the error code would be a breaking change, as we and other implementers might rely on the current code.

PR #954, opened 2024-07-10
chore: request hashicorp registry owner change

apricote:
The registry entry is currently owned by a former maintainer of the provider. We would like to change it to the teams bot account so everyone can access the registry entry if necessary.


jooola (approved):
Must the file stay in the repo forever ?

apricote:
Nope, only until the request was processed. Pretty weird, but if thats what they require :shrug:

PR #982, opened 2024-09-05
refactor: add deprecation mixin for plugin framework

jooola:
Add a deprecation mixin to be used by resource/datasource using the plugin framework.


internal/deprecation/deprecation_plugin.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/hetznercloud/hcloud-go/hcloud"
+)
+
+type Model struct {
+	IsDeprecated         types.Bool   `tfsdk:"is_deprecated"`
+	DeprecationAnnounced types.String `tfsdk:"deprecation_announced"`
+	UnavailableAfter     types.String `tfsdk:"unavailable_after"`
+}
+
+type DeprecationModel = Model //nolint:revive
```
apricote:
```suggestion
// DeprecationModel can be embedded into other resources that implement the Deprecation Info pattern.
// Its name starts with "Deprecation..." so the embedded struct field has a useful name.
type DeprecationModel = Model //nolint:revive
```

internal/deprecation/deprecation_plugin.go
```diff
@@ -0,0 +1,81 @@
...
+
+func NewModel(_ context.Context, in hcloud.Deprecatable) (Model, diag.Diagnostics) {
```
apricote: Not sure this works with `type DeprecationModel = Model`, IIUC Go considers these two to be distinct types, so you can not easily assign values of one type to the other.

I might have misunderstood Go though.

apricote (approved):

internal/deprecation/deprecation_plugin.go
jooola: Seem like a type alias is different from creating a new type (`with =`)

https://go.dev/play/p/zzRD0mYUl-e

```
package main

import "fmt"

type Model struct {
	ID int
}

type DeprecationModel = Model
type DeprecationModelType Model

func main() {
	var value1 Model = DeprecationModel{}
	var value2 Model = DeprecationModelType{}

	var alias1 DeprecationModel = Model{}
	var alias2 DeprecationModelType = Model{}

	fmt.Println(value1)
	fmt.Println(value2)
	fmt.Println(alias1)
	fmt.Println(alias2)
}
```

apricote: TIL!

PR #984, opened 2024-09-05
fix: missing deprecation info in server type list data source

jooola:
The deprecation info were missing from the server types list data source.


internal/servertype/data_source_test.go
```diff
@@ -72,6 +84,19 @@ func TestAccHcloudDataSourceServerTypesTest(t *testing.T) {
 					resource.TestCheckResourceAttr(servertypesD.TFID(), "names.1", "cx21"),
 					resource.TestCheckResourceAttr(servertypesD.TFID(), "descriptions.0", "CX11"),
 					resource.TestCheckResourceAttr(servertypesD.TFID(), "descriptions.1", "CX21"),
```
apricote: These tests fail since today, as the first generation of cx server types was removed from the API.

apricote (approved):

PR #986, opened 2024-09-05
refactor: move control and merge package to util

jooola:
Cleanup the `internal/*` folder.


apricote (approved):
I do not really see the point of this PR. Having properly named packages (which i think `merge` and `control` are) is way nicer than a generic `util` package.

If you think this is better I will not stand in your way though :)

jooola:
I am trying to bring some hierarchy between all our packages, right now we somewhat a flat hierarchy and looking for a resource package, a util package or a mixin takes a bit of time.

I'll keep the smaller packages then;

apricote (approved):

PR #991, opened 2024-09-06
feat: add data.load_balancer_type(s) data sources

### LINKED ISSUE #945: [Feature]: add data source `load_balancer_types`

I would like to see a data source to read all existing lb types from the API (like it is already implemented for locations, datacenters and servers).

I don't think there is much more that can be said about this and no further examples can be given. The data sources mentioned above serve as an example.

jooola:
Closes https://github.com/hetznercloud/terraform-provider-hcloud/issues/945

### DIFF
--- a/internal/loadbalancertype/data_source.go
+++ b/internal/loadbalancertype/data_source.go
@@ -1,0 +1,295 @@
+package loadbalancertype
+
+import (
+	"context"
+	_ "embed"
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/datasourcevalidator"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+
+	"github.com/hetznercloud/hcloud-go/hcloud"
+	"github.com/hetznercloud/terraform-provider-hcloud/internal/util/datasourceutil"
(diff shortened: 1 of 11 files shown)


jooola:
This data source is almost finished, it is missing the following:
- the hcloud-go library does not provide the `deprecated` string field for the load balancer type https://docs.hetzner.cloud/#load-balancer-types-get-a-load-balancer-type, and I am not sure if we want to wait for the backend to implement the new deprecation object before we at it here.
- the documentation is written, but I am not sure if I missed anything.
- I am not sure what the `-x` suffix is used for in the docs sidebar `sidebar_current` template?

jooola:
> the hcloud-go library does not provide the deprecated string field for the load balancer type https://docs.hetzner.cloud/#load-balancer-types-get-a-load-balancer-type, and I am not sure if we want to wait for the backend to implement the new deprecation object before we at it here.

I think we should not implement an attribute that will probably be deprecated soon. I'd therefore merge this without the deprecation information.

> the documentation is written, but I am not sure if I missed anything.

Everything should be written.

> I am not sure what the -x suffix is used for in the docs sidebar sidebar_current template?

Unrelated to this PR.


This PR is ready for review.

internal/loadbalancertype/data_source.go
```diff
@@ -0,0 +1,295 @@
...
+		"id": schema.Int64Attribute{
+			Optional: true,
+			Computed: true,
+		},
+		"name": schema.StringAttribute{
+			Optional: true,
+			Computed: true,
+		},
+		"description": schema.StringAttribute{
+			Computed: true,
```
apricote, line 75: Why is `description` the only non-optional attribute?

apricote (approved):

internal/loadbalancertype/data_source.go
jooola: All the fields except id and name should be read only, I updated this.

apricote (approved):

PR #995, opened 2024-09-11
test: update server type list index to ignore cx11 entry

apricote:
Do not hardcode the list indices of the two server types we verify. Instead get the list from the API and calculate the expected indices.

This makes us more resilient against changes to the list when server types are deprecated/removed.

### DIFF
--- a/internal/servertype/data_source_test.go
+++ b/internal/servertype/data_source_test.go
@@ -1,12 +1,18 @@
 package servertype_test
 
 import (
+	"context"
+	"fmt"
+	"os"
+	"slices"
 	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hetznercloud/hcloud-go/hcloud"
 
 	"github.com/hetznercloud/terraform-provider-hcloud/internal/servertype"
 	"github.com/hetznercloud/terraform-provider-hcloud/internal/teste2e"
-
-	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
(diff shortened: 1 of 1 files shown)


jooola:
Could we update the test data to use cx11 instead? And fix it again once we removed it?

apricote:
> Could we update the test data to use cx11 instead? And fix it again once we removed it?

I updated the PR to instead get the indices of `cpx11` and `cpx21` and use those instead of hardcoding `0` and `1`. What do you think about that approach?

internal/servertype/data_source_test.go
```diff
@@ -67,6 +73,28 @@ func TestAccDataSourceList(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+	serverTypes, err := client.ServerType.All(context.Background())
+	if err != nil {
+		t.Fatalf("failed to get list of server types: %v", err)
+	}
+
+	indexCPX11 := slices.IndexFunc(serverTypes, func(serverType *hcloud.ServerType) bool { return serverType.Name == "cpx11" })
+	indexCPX21 := slices.IndexFunc(serverTypes, func(serverType *hcloud.ServerType) bool { return serverType.Name == "cpx21" })
```
jooola: I think this is overkill, if you really want to keep this, please extract it in a function so we don't make this test more complex that it needs to be.

jooola: I think updating the code once cx11 is really out is better.

jooola:
Let me know what you think

jooola (approved):

apricote:
Works for me!

PR #1018, opened 2024-10-25
docs: add link to locations and datacenters docs

ghostdevv:
Adds links to the upstream Hetzner Docs to give more details about locations and datacenters.

### DIFF
--- a/website/docs/d/datacenter.html.md
+++ b/website/docs/d/datacenter.html.md
@@ -26,6 +26,6 @@ data "hcloud_datacenter" "ds_2" {
 - `id` - (int) Unique ID of the datacenter.
 - `name` - (string) Name of the datacenter.
 - `description` - (string) Description of the datacenter.
-- `location` - (map) Physical datacenter location.
+- `location` - (map) Location details of the datacenter. See the [Hetzner Docs](https://docs.hetzner.com/cloud/general/locations/#what-locations-are-there) for more details about locations.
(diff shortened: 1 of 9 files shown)


jooola:
Hey, 
I'd rather not have the datacenter list hard-coded in the documentation. The API provides datacenters as a dynamic resource, and we might forget to update the docs if a new location is added in the future.

Thanks for the proposal.

ghostdevv:
@jooola there was already a list added to one of these pages in https://github.com/ghostdevv/terraform-provider-hcloud/commit/f2413a6edde0bb5f1b586ab6a7d5197f425ba429, but it was initially confusing to me why it was only in one place and what the `datacenter` string meant. Maybe it is worth an additional page that can be linked to that explains what it is, why it's different to the value of `location`, and what the current values are and how to get the latest ones. I had to go digging through issues to find the command, and I still don't completely understand what the difference in these two fields is.

jooola:
@ghostdevv Hmm, I'll take a closer a look at this. Thanks for bringing this up.

jooola:
@ghostdevv I published a list of available datacenters in the Hetzner Documentation, and added links to it to provide more details if needed.

This way we can safely update our docs once and not have to sync everything in the terraform docs.

Thanks for pushing to improve this!

jooola (approved):

jooola:
Unrelated e2e tests are failing because the PR originates from a fork.

ghostdevv:
Amazing, ty!

PR #1020, opened 2024-10-28
#1019 remove unnecessary start/stop

samene:
Fixes #1019

Issue #550, opened 2022-07-19
ssh connection through the bastion host doesn't work?


medicol69:
### What happened?

I'm using the newest terraform-provider-hcloud version **v1.35.0**, and terraform version **v1.0.7**

I'm trying to create a rancher cluster using only the private ip's because I don't need them and the workers shouldn't be accessed from the internet. So I created a bastion host (with a public ip ofc) but it can't connect to the private interfaces of the worker nodes through ssh??

 I always get this error when using Terraform:

```
│ Error: remote-exec provisioner error
│
│   with module.cluster_init.hcloud_server.rancher_management_nodes[1],
│   on module-cluster-init/resources_hetzner.tf line 74, in resource "hcloud_server" "rancher_management_nodes":
│   74:   provisioner "remote-exec" {
│
│ timeout - last error: ssh: rejected: connect failed (Connection refused)
```
But it all works if I use public ips??

tnx,
Tom

### What did you expect to happen?

I expect that the connection through the bastion host would be able to ssh into the private instances (their private interfaces) of the worker nodes. Like its shown in this example https://www.terraform.io/language/resources/provisioners/connection

### Please provide a minimal working example

```
provisioner "remote-exec" {
    inline = ["cloud-init status --wait > /dev/null"]
    connection {
      type        = "ssh"
      host        = var.servers[count.index]
      user        = "root"
      private_key = file(var.hcloud_ssh_key_path)

      bastion_user        = "root"
      bastion_private_key = file(var.hcloud_ssh_key_path)
      bastion_host = hcloud_server.bastion_host.ipv4_address
    }
  }
```

4ND3R50N:
Hey @medicol69 

one question: 
**are/is your bastion host & your worker nodes all in the same network?** 
Because servers that have no public ips (primary ips) **has to be in a network** to be turned on & running, because theire just living / communicating in the hetzner cloud network. So, you could add your bastion server to the same network (ofc with public ip so you can reach it) to only use internal IPs ([checkout the network block](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/server#server-creation-with-network)) for server -> server communication.


If you still have problems or questions (or maybe i just misunderstood you), just leave a message here (Maybe provide also the server / network configuration)

Have a nice evening :-)

medicol69:
[...]

PR #1021, opened 2024-10-28
fix: remove server network interface workaround

### LINKED ISSUE #1019: [Bug]: server starts and stops two times during first boot
### What happened?

When creating a `hcloud_server` resource the server starts and stops two times before finally starting up. The sequence of events is as below

| activity_type| created | status |triggered_by|
| --------        | ------- |  -------|  -----------|
|server.start | 2024-10-28T05:09:16Z|success|System
|server.start | 2024-10-28T05:09:06Z|requested|BUCw8I***
| server.stop | 2024-10-28T05:09:04Z |success|System
|server.stop | 2024-10-28T05:08:49Z|requested|BUCw8I***
|server.start | 2024-10-28T05:08:43Z|success|System
|server.start | 2024-10-28T05:08:33Z|requested|BUCw8I***
|network.attach | 2024-10-28T05:08:31Z|success|System
|network.attach | 2024-10-28T05:08:29Z|requested|BUCw8I***
|server.create | 2024-10-28T05:08:27Z|success|System
|server.create | 2024-10-28T05:08:24Z|requested|BUCw8I***

When the same server is created from UI or from the official ansible playbook, only the following events are generated
| activity_type| created | status |project|triggered_by
| --------        | ------- |  -------|  -----------|--------|
|network.attach | 2024-10-28T05:05:51Z|success|work|System
|server.create | 2024-10-28T05:05:45Z|requested||Sa***
|network.attach | 2024-10-28T05:05:42Z|success|work|Sa***
|server.create | 2024-10-28T05:05:42Z|requested||Sa***

Why is the server starting and stopping twice? It should only start once.

I believe this is messing up my cloud-init script. Especially the part where I am adding sudo users

```yaml
# auth section
disable_root: false
ssh_pwauth: yes
users:
  - name: cloud-user
    groups: users, admin
    expiredate: '2032-09-01'
    lock_passwd: false
    ssh_authorized_keys:
      - ${public_key}
    sudo: "ALL=(ALL) NOPASSWD:ALL"
    shell: /bin/bash
```
The cloud-init runs the first time, but before it reaches this part the server is restarted. When it starts again and cloud-init runs again but skips the part about sudo user config. Thus my sudo user is missing

Why does the server start and stop twice?

```
resource "hcloud_server" "rancher_vms" {
  count       = var.rancher_nodes
  name        = format("%s-rancher-%d",var.prefix, count.index + 1)
  image       = var.rancher_image
  server_type = var.rancher_flavor
  datacenter  = var.datacenter

  ssh_keys = [data.hcloud_ssh_key.hcloud-pkey.id]
  network {
    network_id = data.hcloud_network.pe-network.id
    alias_ips = []
  }
  public_net {
    ipv4_enabled = false
    ipv6_enabled = false
  }
  user_data = templatefile("user-data.yml.tftpl", {
    http_proxy = var.proxy,
    public_key = data.hcloud_ssh_key.hcloud-pkey.public_key,
    hostname   = format("%s-rancher-%d",var.prefix, count.index + 1)
  })
}
```

The same observation without `user_data` too.

jooola:
Remove a server workaround that was used to fix a private network issue.

The problem that this workaround tried to solve was a timing issue. When the workaround was implemented, the action returned by the `attach to network` endpoint was marked as success without waiting for the network definition to be applied on the host. Today, the actions returned by the `attach to network` endpoint do wait for the network definition to be applied on the host.

The workaround can therefor be removed with a certain confidence.

Supersede #1020 

Fixes #1019

Related to #552 and #550


apricote:
This issue that caused the workaround only happens sporadically, a single successful test run does not give me confidence that the issue went away.

IIRC we tried to setup some e2e tests that were able to reproduce a few months ago with a bastion server and SSH, not sure if we even merged that.

samene:
> This issue that caused the workaround only happens sporadically, a single successful test run does not give me confidence that the issue went away.
> 
> IIRC we tried to setup some e2e tests that were able to reproduce a few months ago with a bastion server and SSH, not sure if we even merged that.

But the workaround causes issues with userdata scripts. If we stop a server which is midway running cloud-init it can cause uncertain behaviour. some scripts do not run again on the next boot.

apricote:
I understand that you have an issue and we should fix that.

I just wanted to say that the suggested solution of removing the workaround is also not without downsides, as it might introduce another bug.

@jooola We previously tried to remove this workaround in the `plugin-framework-server` branch and it looks like we at least finished writing the test case that validates the behavior the workaround originally tried to fix: https://github.com/hetznercloud/terraform-provider-hcloud/commit/3c16a22c43669dacc54f8b8c82409db0d5db0d74

jooola:
Marking as ready to review, only to run the e2e.

jooola:
We found the origin of the workaround. I am confident that we can remove it without breaking anything.

apricote (approved):

apricote:
Thank you @jooola for researching the original conditions that lead to the workaround and making sure the underlying problem is now fixed.

PR #1025, opened 2024-11-06
test: update server type list index after cx11 removal

jooola:
Partially reverts #995

PR #1026, opened 2024-11-06
test: use test location and datacenter variables

jooola:
We sometimes have to update the location of our resources to work around possible shortages.

The variables declared in GitHub were not forwarded to our tests.


jooola:
The e2e fails because we are missing #1025

PR #1028, opened 2024-11-06
chore: add pre-commit setup

jooola:
This adds a pre-commit config and CI job to enforce the codebase to comply with the pre-commit hooks.

There is a big formatting commit.


website/docs/d/volumes.html.md
```diff
@@ -7,10 +7,11 @@ Provides details about multiple Hetzner Cloud volumes.
 ---
 
 # hcloud_volumes
-Provides details about multiple Hetzner Cloud volumes.
 
+Provides details about multiple Hetzner Cloud volumes.
```
lukasmetzner: Super small nitpick here. In the other docs you are usually writting "Hetzner Cloud Servers" or "Hetzner Cloud Networks" (so title case), but here you have "Hetzner Cloud **v**olumes". I would propose "Hetzner Cloud **V**olumes" to be uniform. But maybe this was already there previously.

lukasmetzner (changes requested):

website/docs/d/volumes.html.md
jooola: I was there already, but agreed. I fixed it.

lukasmetzner (approved):

PR #1030, opened 2024-11-06
feat: drop support for opentofu v1.6

jooola:
Removes support for opentofu v1.6.

See https://endoflife.date/opentofu

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,8 +49,6 @@ jobs:
       matrix:
         include:
           - tool: opentofu
-            version: v1.6.x
-          - tool: opentofu
             version: v1.7.x
           - tool: terraform
             version: v1.8.x


lukasmetzner (approved):

ghost:
Hey folks, the information on the endoflife.date website is incorrect (see endoflife-date/endoflife.date#6203 ). OpenTofu 1.6 is still supported at least until 1.9 comes out.

jooola:
Hmm, yes indeed! 

@abstractionfactory The support removal is, essentially, removing the version from our test matrix, so you should still be able to us it.

I see that opentofu has already released a [v1.9.0-alpha1](https://github.com/opentofu/opentofu/releases/tag/v1.9.0-alpha1), so I assume we can expect a 1.9.0 release in the coming month.

With that in mind, would you be ok if we did not revert this PR and simply wait for the next 1.9.0 release?

cc @apricote

ghost:
@jooola that is correct, 1.9 is coming in a few weeks. I think we still need to have a conversation with the TSC if we want to extend support for 1.6 in order to help folks who are migrating.

Nevertheless, the only change we made in 1.7 related to providers are the provider-defined functions, which is also new code, so I think your current CI setup should continue to serve you well even if we extend 1.6 support.

Feel free to reach out on the OpenTofu Slack if you would like to talk more about this topic, I just wanted to make sure you are aware that the data on endoflife.date is incorrect. :)

PR #1040, opened 2024-11-14
refactor: move e2e tests into own packages

phm07:
This refactor moves the e2e tests into their own packages so they can be tested separately from regular unit tests.
Also tests have been nested using Go's `t.Run` to make the test tree more legible and organized.


jooola:
This should already be possible, if you don't define TF_ACC=1, the e2e test should not run.

I prefer not to move the tests files too far from the sources.

If this is really needed, we can use a build tag.

phm07:
> This should already be possible, if you don't define TF_ACC=1, the e2e test should not run.

The goal here is to be able to run the e2e tests on their own, without the unit tests

Issue #877, opened 2024-02-23
[Feature]: Update to hcloud-go v2


apricote:
### What whould you like to see?

We should switch from the deprecated hcloud-go v1 to v2. This will be possible once we have fully migrated to the new Plugin Framework (#752).

jooola:
Solved by #1062

PR #1062, opened 2024-12-23
fix: upgrade hcloud-go to v2

jooola:
Attempt to migrate the hcloud-go library to v2, to be able to finally drop the v1 version.

We previously discussed of not going this route, but the effort of porting new features in both hcloud-go versions is bigger than upgrade the library in the terrafrom plugin.

If this change ends up being impossible to test and too dangerous, we will just continue the route we initially choose: migrating for to the plugin framework then the hcloud-go library.

Related to #877


internal/util/resourceutil/id.go
```diff
@@ -27,9 +27,9 @@ func ParseID(value types.String) (int, diag.Diagnostics) {
 			"ID is larger than your system supports.",
 			"The current version of the provider has problems with IDs > 32bit on 32 bit systems. If possible, switch to a 64 bit system for now. See https://github.com/hetznercloud/hcloud-go/issues/263")
 	}
-	return int(id), diagnostics
+	return id, diagnostics
```
apricote, line 25: The check above this line (`id > math.MaxInt32 && strconv.IntSize == 32`) can be removed, it was only relevant for the `int(id)` cast.

internal/util/id.go
```diff
@@ -0,0 +1,25 @@
...
+	case int64:
+		return o
+	case int:
+		return int64(o)
+	default:
+		panic(fmt.Sprintf("unexpected type %T for %#v", v, v))
+	}
+}
+
+func FormatID[T ~int | ~int32 | ~int64](v T) string {
```
apricote: Is `~int32` actually used anywhere? This would always result in a runtime panic in `CastInt64()`, so maybe returning a compile-time error instead would be easier to spot?

internal/util/id.go
```diff
@@ -0,0 +1,25 @@
...
+	return strconv.FormatInt(CastInt64(v), 10)
+}
+
+func ParseID(v string) (int64, error) {
+	return strconv.ParseInt(v, 10, 32)
```
apricote: Not sure where this is used yet, but why limit it to `32` bits?

apricote (commented):
Looking through it I am hopeful for the future of `hcloud-go`!

I am a bit worried about the naked `int()` castings to downcast for terraform on 32 bit platforms. Go just does it, resulting in bad data. What do you think about adding a reverse function that converts from int64 to int on 64-bit platforms, and errors on 32-bit platforms?

jooola:
> I am a bit worried about the naked int() castings to downcast for terraform on 32 bit platforms. Go just does it, resulting in bad data. What do you think about adding a reverse function that converts from int64 to int on 64-bit platforms, and errors on 32-bit platforms?

Yup, that's a sensible idea. I addressed your review comments and implement this for the relevant casts.

apricote (approved):
Very nice :rocket: This looks good as far as I can tell, happy to finally put hcloud-go v1 to bed :)

PR #1066, opened 2025-01-09
feat: drop support for terraform v1.8

jooola:
Removes support for terraform v1.8

See https://endoflife.date/terraform


lukasmetzner (approved):

lukasmetzner:
Any reason why we keeping 1.7 though?

jooola:
> Any reason why we keeping 1.7 though?

1.7 is the opentofu version :)

PR #1089, opened 2025-02-19
test: use sprig template function

jooola:
This provide use many more functions to template the test tf files.


jooola:
All CI failures are from the API resource shortage. The templates is therefor working as expected.

apricote (approved):

Issue #1048, opened 2024-11-21
[Bug]: Private network not included on server data source


inakimalerba:
### What happened?

When trying to fetch information about an existing server, for example, by using the following query, the resulting objects do not contain information about the private networks.

```
data "hcloud_servers" "masters" {
  with_selector = "role=master"
}
```

I think the issue is on [this check](https://github.com/hetznercloud/terraform-provider-hcloud/blob/main/internal/server/resource.go#L1189), which apparently does not work when we're fetching the server information: 

Also, the [data schema](https://github.com/hetznercloud/terraform-provider-hcloud/blob/main/internal/server/data_source.go#L25) does not contain a network field.

### What did you expect to happen?

Be able to get the value of the private IP of a VM.

### Please provide a minimal working example

```
data "hcloud_servers" "masters" {
  with_selector = "role=master"
}

output "masters" {
  value = data.hcloud_server.masters
}
```

output is

```
   masters = {
       id            = "redacted"
       servers       = [
           {
               backup_window      = ""
               backups            = false
               datacenter         = "nbg1-dc3"
               delete_protection  = false
               firewall_ids       = []
               id                 = redacted
               image              = "debian-12"
               ipv4_address       = ""
               ipv6_address       = ""
               ipv6_network       = "<nil>"
               iso                = ""
               labels             = {
                   cluster = "staging"
                   role    = "master"
                }
               location           = "nbg1"
               name               = "staging-master1"
               network            = []
               placement_group_id = redacted
               primary_disk_size  = 80
               rebuild_protection = false
               rescue             = ""
               server_type        = "cpx21"
               status             = "running"
            },
            ...
``

castarco:
In theory, according to documentation ( https://search.opentofu.org/provider/opentofu/hcloud/latest/docs/datasources/server#attributes-reference ), and according to this PR ( https://github.com/hetznercloud/terraform-provider-hcloud/pull/1115 )... this issue should be already solved.

Perhaps this issue can be marked as fixed?

> **Part of what I wrote before that is no longer relevant:**
> 
> ~However... that's not really the case; at least when the `server` data source is obtained from a `servers` data source.~
> 
> I'm having the following error:
> ```
> │ Error: Unsupported attribute
> │ 
> │   on outputs.tf line 32, in output "ks_prod_servers_info_for_ansible":
> │   32:     vpc_ip: s.network.ip
> │     ├────────────────
> │     │ s.network is set of object with 1 element
[...]

PR #1115, opened 2025-04-11
feat(server): add private networks in data_source

xavierleune:
Hi hetzner,

This provider is super cool ! I noticed something that would be a nice to have: the current data_source for servers does not outputs the private network data. That would be a great addition, as it makes working with private-network-only servers easier. There is a use case identified around autoscaling nodes [here](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner/pull/1567#discussion_r2039499973).
I'm not sure if that's the best way to go, but that's a starting point, feel free to give feedback,

Have a nice week-end,

### DIFF
--- a/internal/server/data_source.go
+++ b/internal/server/data_source.go
@@ -91,6 +91,31 @@ func getCommonDataSchema() map[string]*schema.Schema {
 			Computed: true,
 			Elem:     &schema.Schema{Type: schema.TypeInt},
 		},
+		"network": {
+			Type:     schema.TypeSet,
+			Optional: true,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"network_id": {
+						Type:     schema.TypeInt,
+						Computed: true,
+					},
+					"ip": {
+						Type:     schema.TypeString,
+						Computed: true,
+					},
+					"alias_ips": {
+						Type:     schema.TypeSet,
+						Elem:     &schema.Schema{Type: schema.TypeString},
+						Computed: true,
+					},
+					"mac_address": {
+						Type:     schema.TypeString,
+						Computed: true,
+					},
+				},
+			},
+		},
 		"placement_group_id": {
 			Type:     schema.TypeInt,
 			Optional: true,
@@ -167,6 +192,13 @@ func DataSourceList() *schema.Resource {
 	}
 }
 
+func dataSourceServerItem() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceHcloudServerListRead,
+		Schema:      getCommonDataSchema(),
+	}
+}
+
 func dataSourceHcloudServerRead(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 	client := m.(*hcloud.Client)
 
@@ -178,7 +210,7 @@ func dataSourceHcloudServerRead(ctx context.Context, d *schema.ResourceData, m i
 		if s == nil {
 			return diag.Errorf("no Server found with id %d", id)
 		}
-		setServerSchema(d, s)
+		setServerSchema(d, s, true)
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


xavierleune:
Related issue: #1048

internal/server/data_source.go
```diff
@@ -91,6 +91,33 @@ func getCommonDataSchema() map[string]*schema.Schema {
...
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"network_id": {
+						Type:     schema.TypeInt,
+						Required: true,
+					},
+					"ip": {
+						Type:     schema.TypeString,
+						Computed: true,
+						Optional: true,
```
apricote: For the data source all fields are always set, so you can remove the `Optional: true` from `ip`, `alias_ips` and `mac_address` here and in the docs.

internal/server/data_source.go
```diff
@@ -91,6 +91,33 @@ func getCommonDataSchema() map[string]*schema.Schema {
...
 			Elem:     &schema.Schema{Type: schema.TypeInt},
 		},
+		"network": {
+			Type:     schema.TypeSet,
+			Optional: true,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"network_id": {
+						Type:     schema.TypeInt,
+						Required: true,
```
apricote: This field should be `Computed: true` instead of required, as the user can not specify it but the provider sets it.

internal/server/resource.go
```diff
@@ -1126,11 +1126,11 @@ func newIPSet(f schema.SchemaSetFunc, ips []net.IP) *schema.Set {
...
 }
 
-func setServerSchema(d *schema.ResourceData, s *hcloud.Server) {
-	util.SetSchemaFromAttributes(d, getServerAttributes(d, s))
+func setServerSchema(d *schema.ResourceData, s *hcloud.Server, data_source bool) {
+	util.SetSchemaFromAttributes(d, getServerAttributes(d, s, data_source))
 }
 
-func getServerAttributes(d *schema.ResourceData, s *hcloud.Server) map[string]interface{} {
+func getServerAttributes(d *schema.ResourceData, s *hcloud.Server, data_source bool) map[string]interface{} {
```
apricote: Could you rename this to `forceSetNetworkAttribute` instead of `data_source` in this method and in `setServerSchema`?

templates/data-sources/server.md.tmpl
```diff
@@ -36,8 +36,16 @@ This resource is useful if you want to use a non-terraform managed server.
...
 - `delete_protection` - (bool) Whether delete protection is enabled.
 - `rebuild_protection` - (bool) Whether rebuild protection is enabled.
 - `primary_disk_size` - (int) The size of the primary disk in GB.
+
+a single entry in `network` support the following fields:
+
+- `network_id` - (Required, int) ID of the network
+- `ip` - (Optional, string) Specify the IP the server should get in the network
+- `alias_ips` - (Optional, list) Alias IPs the server should have in the Network.
+- `mac_address` - (Optional, string) The MAC address the private interface of the server has
```
apricote: Users can not set these fields, so "[field] the server should get/have in the network" does not really make sense. Please update it to indicate that this is what the server **has**.

apricote (changes requested):
Nice!

Please also add an e2e test for this scenario or add it to an [existing e2e test](https://github.com/hetznercloud/terraform-provider-hcloud/blob/6988c3299c8fe717575c3b90898e76745e7cbe7e/internal/server/data_source_test.go#L18).

xavierleune:
Hi @apricote
Thanks for your feedback, I'm not quite sure if it's ok for e2e tests, you may wanna double check that ^^

xavierleune:
@apricote could you please review this pull request? Do you see any other changes that need to be made? Thanks !

apricote:
Hey @xavierleune,

sorry for the delay. I added a test for this in https://github.com/hetznercloud/terraform-provider-hcloud/pull/1115/commits/3917fa1023fbd4936b1845af951fa1777f67d573, feel free to check it out if you want to learn more about it.

I also ran the e2e test suite locally and everything looks good :)

apricote (approved):

xavierleune:
@apricote thanks for the merge and for the tests 🙏

PR #1116, opened 2025-04-14
feat: add support for providing volume at the server creation time

### LINKED ISSUE #502: [Feature]: Allow creation of server with volume(s)

`hcloud_server` does not allow to attach directly volumes when the server is created, but the API does as desribed in the [docs](https://docs.hetzner.cloud/#servers-create-a-server).

Can you please support this? Or is there another way to do this? Attaching after creation would not work in my case.

Thanks.

kirillmakhonin:
This closes #502 and adds ability to provide ID of volumes at the time of server creation.

### DIFF
--- a/internal/server/resource.go
+++ b/internal/server/resource.go
@@ -103,6 +103,14 @@ func Resource() *schema.Resource {
 					}},
 				ForceNew: true,
 			},
+			"volumes": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				ForceNew: true,
+			},
 			"keep_disk": {
 				Type:     schema.TypeBool,
 				Optional: true,
@@ -331,6 +339,11 @@ func resourceServerCreate(ctx context.Context, d *schema.ResourceData, m interfa
 	}
 
 	opts.SSHKeys, err = getSSHkeys(ctx, c, d)
+	if err != nil {
+		return hcloudutil.ErrorToDiag(err)
+	}
+
+	opts.Volumes, err = getVolumes(ctx, c, d)
 	if err != nil {
 		return hcloudutil.ErrorToDiag(err)
 	}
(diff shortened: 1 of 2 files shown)


kirillmakhonin:
@jooola  please review

internal/server/resource.go
```diff
@@ -103,6 +103,14 @@ func Resource() *schema.Resource {
 					}},
 				ForceNew: true,
 			},
+			"volumes": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				ForceNew: true,
```
apricote: Can you explain why this forces a new resource? It is possible to attach or detach volumes when the server already exists.

If this is because some of the usecases would benefit from this, ie. provisioning with cloud-init that needs to re-run if the volumes change, then this should be in the hands of the user, and not forced by the provider. There is an open TF issue for this: https://github.com/hashicorp/terraform/issues/27456 Not sure if its possible to reference the resources own fields in the existing [`lifecycle.replace_triggered_by`](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#replace_triggered_by) field.

apricote (changes requested):
Please also add the new field to our [documentation](https://github.com/hetznercloud/terraform-provider-hcloud/blob/main/docs/resources/server.md) and add an [end-to-end test](https://github.com/hetznercloud/terraform-provider-hcloud/blob/main/internal/server/resource_test.go) for this field.

kirillmakhonin:
@apricote thank you for feedback
Yes, main goal of this PR is to allow creation of servers with volumes mounted at the first boot time so cloud-init has access to volumes.
I've updated this PR to address your comments, but only I'm not sure how to address comment with testing - was not able to find any documentation regarding creation of tests. Is it something you can help with? I believe it is only one blocker to add this feature which people (including me) are waiting to be able to use Hetzner Cloud for more advanced use cases with IaaC

jooola:
Sorry, we haven't had time to give a proper review. Reopening the PR to make sure we come back to it.

kirillmakhonin:
Hello
Any update on this?

jooola:
As already stated in the initial review, please: 
- Add the new field to our [documentation](https://github.com/hetznercloud/terraform-provider-hcloud/blob/main/docs/resources/server.md), only adding an example is not enough.
- Add an [end-to-end test](https://github.com/hetznercloud/terraform-provider-hcloud/blob/main/internal/server/resource_test.go) for this field. The field must be tested when creating the resource and when updating the resource.

jooola:
Closing as stalled.

PR #1120, opened 2025-05-02
chore: clean up from #1115

apricote:
I missed these two issues when reviewing and merging the PR. This should fix CI in main.


docs/data-sources/server.md
```diff
@@ -46,8 +46,16 @@ data "hcloud_server" "s_3" {
 - `ipv6_network` - (string) The IPv6 network.
 - `status` - (string) The status of the server.
 - `labels` - (map) User-defined labels (key-value pairs)
+- `network` - (map) Private Network the server is attached to.
```
jooola, line 49: I think this is the wrong file, you might need to use the docs in the templates dir.

jooola: Nevermind.

jooola (approved):

PR #1133, opened 2025-06-03
refactor: check if labels are unknown

jooola:
When converting terraform labels to API labels, make sure we account for unknown labels.


apricote (approved):

jooola:
I am not 100% sure about this change yet.

jooola:
I am more confident with the test I added.

PR #1145, opened 2025-06-28
feat(871): add support and documentation for ISO datasources

mekahell:
Add the following datasource for ISO:
* `hcloud_iso` : to retrieve an ISO from its id, name, architecture or name prefix
* `hcloud_isos` : to retrieve a list of ISO from their name, architecture or name prefix

The documentation has been updated with these 2.

### DIFF
--- a/docs/data-sources/iso.md
+++ b/docs/data-sources/iso.md
@@ -1,0 +1,53 @@
+---
+page_title: "Hetzner Cloud: hcloud_iso"
+description: |-
+  Provides details about a specific Hetzner Cloud ISO.
+---
+
+# Data Source: hcloud_iso
+
+Provides details about a Hetzner Cloud ISO.
+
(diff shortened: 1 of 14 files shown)


jooola:
Hey @mekahell  Thanks for the contribution!

We are in the process of migrating our resources and data sources to the Terraform plugin framework (see #752). The implementation of the ISO data sources are using the older pattern.

Would you have time to convert them to use the Terraform plugin framework? You can find an example data source in https://github.com/hetznercloud/terraform-provider-hcloud/tree/main/internal/location

apricote:
Hey @mekahell,

we do not accept new implementations using Terraform SDKv2. If you (or anyone else) is interested in implementing the resources using the Plugin Framework, we are more than happy to review and merge them :)

Kind Regards

PR #1167, opened 2025-08-06
feat: add support for Hetzner API endpoint

jooola:
Add the new `endpoint_hetzner` provider option, to override the default `https://api.hetzner.com/v1` API endpoint.


apricote (changes requested):
The changes are looking good, but they are missing an update to the [provider docs](https://github.com/hetznercloud/terraform-provider-hcloud/blob/main/docs/index.md#argument-reference)

apricote (approved):

PR #1191, opened 2025-09-24
fix: add experimental features maturity

jooola:
Add the API upstream maturity information to our experimental features release
process.

### DIFF
--- a/internal/util/experimental/experimental.go
+++ b/internal/util/experimental/experimental.go
@@ -11,7 +11,7 @@ import (
 //
 // Usage:
 //
-//	var Product = New("Product name", "https://docs.hetzner.cloud/changelog#new-product")
+//	var Product = New("Product name", "in beta", "https://docs.hetzner.cloud/changelog#new-product")
 //
 //	func (r *Resource) Configure(_ context.Context, _ resource.ConfigureRequest, resp *resource.ConfigureResponse) {
 //		experimental.Product.AppendDiagnostic(&resp.Diagnostics)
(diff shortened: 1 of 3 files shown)


internal/util/experimental/experimental.go
```diff
@@ -26,28 +26,29 @@ import (
 //		experimental.Product.AppendNotice(&resp.Schema.MarkdownDescription)
 //	}
 type Experimental struct {
-	product string
-	url     string
+	product  string
+	maturity string
```
lukasmetzner, line 30: Should we enum this? I am fine to leave it like this.

lukasmetzner (approved):

internal/util/experimental/experimental.go
```diff
@@ -11,7 +11,7 @@ import (
 //
 // Usage:
 //
-//	var Product = New("Product name", "https://docs.hetzner.cloud/changelog#new-product")
+//	var Product = New("Product name", "beta", "https://docs.hetzner.cloud/changelog#new-product")
```
phm07: This will lead to the following text:
`Experimental: Product name is beta, ...`
We probably want to change the example so it says "in beta" instead

jooola: Works for me

internal/util/experimental/experimental.go
jooola: Maybe, if this makes the suggestion from Paul easier to implement, I'll have a look later

internal/util/experimental/experimental.go
phm07: Should we capitalize it or not? E.g. `beta` vs `Beta`

Issue #672, opened 2023-04-28
[Feature]: hcloud_server_network: assign IP from supplied subnet


apricote:
### What whould you like to see?

When assigning a server to a network through the `hcloud_server_network` resource, it is not possible to authoritavely choose the subnet that will be used. From out [current docs](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/server_network#subnet_id):

> [subnet_id](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/server_network#subnet_id) - (Optional, string) ID of the sub-network which should be added to the Server. Required if network_id is not set. **_Note_: if the ip property is missing, the Server is currently added to the last created subnet.**

This happens because the [API](https://docs.hetzner.cloud/#server-actions-attach-a-server-to-a-network) does not support "attaching to a subnet", it only supports attaching to a Network with an optional IP. This IP would be used to determine the subnet the server is put into.

If a users wants the server to be "put into the subnet", they would need to explicitly calculate an IP from that subnet and pass that to the resource. While this is easy to do in simple setups, it can add dev cost in larger setups.

I am not really sure how we can implement this, besides generating random IPs from the subnet and trying to attach the server as this IP, but this always has the potential to conflict because that IP is already in use. Getting all assigned IPs and just assigning the next one has an even larger potential for conflicts, because that same IP will be chosen by all `hcloud_server_network` resources that are evaluated at the same time.

Rohmilchkaese:
Push for that one. That exactly is the issue I encountered. 

This is the code I use, to assign to private IPs:
```
#Asign private IPs to Master
resource "hcloud_server_network" "private_subnet_master" {
  for_each  = local.master
  server_id = lookup(hcloud_server.master[each.key], "id")
  subnet_id = hcloud_network_subnet.private_subnet_master.id
}

#Asign private IPs to Worker
resource "hcloud_server_network" "private_subnet_worker" {
  for_each  = local.worker
  server_id = lookup(hcloud_server.worker[each.key], "id")
  subnet_id = hcloud_network_subnet.private_subnet_worker.id
}
```

Which in return results in:
<img width="1493" alt="Screenshot 2023-05-01 at 15 10 13" src="https://user-images.githubusercontent.com/37877524/235455997-cf312270-7925-4800-a362-6de976db9e9b.png">
[...]

PR #1217, opened 2025-10-16
feat: attach server to a specific network subnet

jooola:
Allows the users to attach a server to a given network subnet.

Using the `subnet_id` argument is preferred over the `network_id` argument.

### DIFF
--- a/internal/server/resource_network.go
+++ b/internal/server/resource_network.go
@@ -78,7 +78,7 @@ Manage the attachment of a Server in a Network in the Hetzner Cloud.
 			},
 		},
 		"network_id": schema.Int64Attribute{
-			MarkdownDescription: "ID of the Network to attach the Server to. Using `subnet_id` is preferred. Required if `subnet_id` is not set. If `ip` is not set, the Server will be attached to the last subnet (ordered by `ip_range`).",
+			MarkdownDescription: "ID of the Network to attach the Server to. Using `subnet_id` is preferred. Required if `subnet_id` is not set. If `subnet_id` or `ip` are not set, the Server will be attached to the last subnet (ordered by `ip_range`).",
 			Optional:            true,
 			Computed:            true,
 			PlanModifiers: []planmodifier.Int64{
@@ -87,7 +87,7 @@ Manage the attachment of a Server in a Network in the Hetzner Cloud.
 			},
 		},
 		"subnet_id": schema.StringAttribute{
-			MarkdownDescription: "ID of the Subnet to attach the Server to. Required if `network_id` is not set. If `ip` is not set, the Server will be attached to the last subnet (ordered by `ip_range`).",
+			MarkdownDescription: "ID of the Subnet to attach the Server to. Required if `network_id` is not set.",
 			Optional:            true,
 			PlanModifiers: []planmodifier.String{
 				stringplanmodifier.RequiresReplace(),
@@ -154,7 +154,7 @@ func (r *NetworkResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 	}
 
 	if !data.SubnetID.IsUnknown() && !data.SubnetID.IsNull() {
-		subnetNetwork, _, err := r.ParseSubnetID(data.SubnetID.ValueString())
+		subnetNetwork, subnetIPRange, err := r.ParseSubnetID(data.SubnetID.ValueString())
 		if err != nil {
 			resp.Diagnostics.AddAttributeError(
 				path.Root("subnet_id"),
@@ -180,6 +180,28 @@ func (r *NetworkResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 				resp.RequiresReplace.Append(path.Root("network_id"))
 			}
 		}
+
+		if !data.IP.IsUnknown() && !data.IP.IsNull() {
+			// Check if the attachment IP (state) is within the subnet ip range.
+			attachmentIP := net.ParseIP(data.IP.ValueString())
+			if !subnetIPRange.Contains(attachmentIP) {
+				resp.Diagnostics.AddAttributeWarning(
+					path.Root("subnet_id"),
+					"Attachment IP is outside subnet IP range",
+					fmt.Sprintf(
+						"Attachment IP (%s) is outside subnet IP range (%s) (%s).",
+						attachmentIP.String(), subnetIPRange, data.SubnetID.ValueString(),
+					),
+				)
+
+				// Only marking the attribute as "RequiresReplace" does not work, as
+				// terraform core internally filters out any replacements to attributes that
+				// have not changed. Marking the attribute as unknown is correct, and makes
+				// it so the RequiresReplace is actually applied and shown to the user.
+				resp.Diagnostics.Append(resp.Plan.SetAttribute(ctx, path.Root("ip"), iptypes.NewIPAddressUnknown())...)
+				resp.RequiresReplace.Append(path.Root("ip"))
+			}
+		}
 	}
 }
 
@@ -199,7 +221,7 @@ func (r *NetworkResource) Create(ctx context.Context, req resource.CreateRequest
 		opts.Network = &hcloud.Network{ID: data.NetworkID.ValueInt64()}
 	}
 	if !data.SubnetID.IsUnknown() && !data.SubnetID.IsNull() {
-		subnetNetwork, _, err := r.ParseSubnetID(data.SubnetID.ValueString())
+		subnetNetwork, subnetIPRange, err := r.ParseSubnetID(data.SubnetID.ValueString())
 		if err != nil {
 			resp.Diagnostics.AddAttributeError(
 				path.Root("subnet_id"),
@@ -208,6 +230,7 @@ func (r *NetworkResource) Create(ctx context.Context, req resource.CreateRequest
 			)
 		}
 		opts.Network = subnetNetwork
+		opts.IPRange = subnetIPRange
 	}
 
 	if !data.IP.IsUnknown() && !data.IP.IsNull() {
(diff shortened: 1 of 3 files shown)


internal/server/resource_network.go
```diff
@@ -2,355 +2,509 @@ package server
...
+					"Invalid Subnet ID",
+					fmt.Sprintf(
+						"The network ID (%s) and the subnet ID (%s) do not refer to the same network ID.",
+						data.NetworkID,
+						data.SubnetID,
+					),
+				)
+			}
+			opts.Network = subnetNetwork
+			opts.IPRange = subnetIPRange
```
jooola: TODO: remove this line to remove the attach to subnet ip_range feature.

Test might need to be adjusted accordingly

internal/util/strings.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+func TitleCase(s string) string {
+	if s == "" {
+		return ""
+	}
+	return string(unicode.ToTitle(rune(s[0]))) + s[1:]
+}
+
+func Dedent(s string) string {
```
apricote, line 10: This function is unused as far as I can tell.

internal/server/resource_network.go
```diff
@@ -2,355 +2,509 @@ package server
...
-			"network_id": {
-				Type:     schema.TypeInt,
-				Optional: true,
-				ForceNew: true,
+var _ resource.Resource = (*networkResourceImpl)(nil)
+var _ resource.ResourceWithConfigure = (*networkResourceImpl)(nil)
+var _ resource.ResourceWithConfigValidators = (*networkResourceImpl)(nil)
+var _ resource.ResourceWithImportState = (*networkResourceImpl)(nil)
+
+type networkResourceImpl struct {
```
apricote: Is there a reason this resource uses a different naming schema to other plugin framework resources?

```suggestion
type NetworkResource struct {
```

internal/server/resource_network.go
```diff
@@ -2,355 +2,509 @@ package server
...
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
+		},
+		"server_id": schema.Int64Attribute{
+			MarkdownDescription: "ID of the Server.",
+			Required:            true,
+			PlanModifiers: []planmodifier.Int64{
+				int64planmodifier.RequiresReplace(),
+				int64planmodifier.UseStateForUnknown(),
```
apricote: This field is required, so it can never be `unknown`.

```suggestion
```

internal/server/resource_network.go
```diff
@@ -2,355 +2,509 @@ package server
...
-func resourceServerNetworkCreate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
-	client := m.(*hcloud.Client)
-	ip := net.ParseIP(d.Get("ip").(string))
+// ConfigValidators returns a list of functions which will all be performed during validation.
+func (r *networkResourceImpl) ConfigValidators(context.Context) []resource.ConfigValidator {
+	return []resource.ConfigValidator{
+		resourcevalidator.AtLeastOneOf(
+			path.MatchRoot("network_id"),
+			path.MatchRoot("subnet_id"),
+		),
```
apricote: Is it valid to specify both? The previous implementation gave an error in this case (`"either network_id or subnet_id must be specified"`).

Allowing both to be set introduces challenges where they differ, which you handle for `Create`. I think there is a possible bug on import though, see the thread on `Read()`.)

internal/server/resource_network.go
```diff
@@ -2,355 +2,509 @@ package server
...
+	}
+}
 
-	networkID, nwIDSet := d.GetOk("network_id")
+type networkResourceData struct {
```
apricote: Do you plan to move this to `model.go`?

internal/server/resource_network.go
```diff
@@ -2,355 +2,509 @@ package server
...
-	server := &hcloud.Server{ID: util.CastInt64(d.Get("server_id"))}
-	n := &hcloud.Network{ID: util.CastInt64(networkID)}
-	aliasIPs := make([]net.IP, 0, d.Get("alias_ips").(*schema.Set).Len())
-	for _, aliasIP := range d.Get("alias_ips").(*schema.Set).List() {
-		ip := net.ParseIP(aliasIP.(string))
-		aliasIPs = append(aliasIPs, ip)
-	}
+	data.ID = types.StringValue(fmt.Sprintf("%d-%d", server.ID, attachment.Network.ID))
+	data.ServerID = types.Int64Value(server.ID)
+	data.NetworkID = types.Int64Value(attachment.Network.ID)
```
apricote: In the previous implementation we only set `network_id` if `subnet_id` was empty.

Not sure if this actually causes issues.

internal/server/resource_network.go
```diff
@@ -2,355 +2,509 @@ package server
...
 
-	err := attachServerToNetwork(ctx, client, server, n, ip, aliasIPs)
-	if err != nil {
-		return hcloudutil.ErrorToDiag(err)
+	if !data.AliasIPs.IsNull() || len(attachment.Aliases) > 0 {
+		aliasIPsStrings := sliceutil.Transform(attachment.Aliases, func(e net.IP) string { return e.String() })
+
+		data.AliasIPs, newDiags = types.SetValueFrom(ctx, types.StringType, aliasIPsStrings)
+		diags.Append(newDiags...)
 	}
```
apricote: This removes a 5-year-old workaround where the Alias IPs tried to keep a consistent order. I believe the workaround should not be needed as this is a set anyway, where the order should not matter.

Have you looked into the comment in the old implementation?

internal/server/resource_network.go
```diff
@@ -2,355 +2,509 @@ package server
...
+			)
+		}
+		if !subnetIPRange.Contains(attachment.IP) {
+			resp.Diagnostics.AddAttributeWarning(
+				path.Root("subnet_id"),
+				"Assigned IP is outside subnet IP range",
+				"",
+			)
+		}
 	}
```
apricote: I have a feeling there is a bug here:
  
- A resource is imported with ID of network1
- User specified subnet from network2
- During read we only check if the ip range matches

So we may not detect the conflict between `subnet_id` and `network_id`, which is explicitly checked in `Create`.

internal/server/resource_network.go
```diff
@@ -2,355 +2,509 @@ package server
...
-				Computed: true,
-				Optional: true,
-				ForceNew: true,
+		},
+		"subnet_id": schema.StringAttribute{
+			MarkdownDescription: "ID of the Subnet to attach the Server to. Required if `network_id` is not set.",
+			Optional:            true,
+			PlanModifiers: []planmodifier.String{
+				stringplanmodifier.RequiresReplace(),
+				stringplanmodifier.UseStateForUnknown(),
```
apricote: As this is not set to `Computed: true`, there is no way for us to return a value. No need to `UseStateForUnknown`.


```suggestion
```

internal/server/resource_network.go
```diff
@@ -2,355 +2,509 @@ package server
...
 
-	if d.HasChange("alias_ips") {
-		if err := updateServerAliasIPs(ctx, client, server, network, d.Get("alias_ips").(*schema.Set)); err != nil {
-			return hcloudutil.ErrorToDiag(err)
-		}
+	if !data.AliasIPs.IsUnknown() && !data.AliasIPs.IsNull() {
+		aliasIPsRaw := make([]string, 0, len(data.AliasIPs.Elements()))
+		resp.Diagnostics.Append(data.AliasIPs.ElementsAs(ctx, &aliasIPsRaw, false)...)
+
+		opts.AliasIPs = sliceutil.Transform(aliasIPsRaw, net.ParseIP)
```
apricote, line 244: Pre-existing bug: Invalid Alias IPs are set to `nil` and sent to the API that way (probably as `0.0.0.0`)

internal/server/resource_network.go
jooola: It is following the ssh key resource naming pattern. Where we could not use the `resource` name since it is a package.

But we could use `Resource` for sure.

internal/server/resource_network.go
jooola: Changed to exactly one of subnet tor network.

internal/server/resource_network.go
jooola: I did see the comment, but ignored it because tf framework set attributes are "unordered".

https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes/set

internal/server/resource_network.go
jooola: Added an additional check if the network ID is not the same as the attachment network ID.

internal/server/resource_network.go
jooola: We now have a IP validator

jooola:
Depends on #1244

apricote (approved):
Looks good, but requires an entry in the release notes, as the behaviour has changed and `subnet_id` + `ip` is more strongly validated

shinebayar-g:
Hi. Does this solve #672 ? :slightly_smiling_face:

privatecoder:
@apricote @lukasmetzner: @shinebayar-g already asked this in #672 ([here](https://github.com/hetznercloud/terraform-provider-hcloud/issues/672#issuecomment-2616606266)) and I think it is a) important and b) I wouldn’t know where else I should write or post this... So:

Given that subnet-support was recently introduced to autoscaler along with an API change I would kindly ask to also consider adding this feature to [server creation](https://docs.hetzner.cloud/reference/cloud#tag/servers/create_server) as well.

This is especially important as currently when selecting a private network (without any public IP) during creation of an instance **I cannot change the subnet afterwards** as this would leave the instance without any IP during the _swap_.

So in order achieve this, it currently requires the user to book (and pay for) a public IP in order to disconnect from the assigned subnet and re-assign it to the desired one.

PR #1225, opened 2025-10-21
feat: support managing records of type SOA

### LINKED ISSUE #1222: [Feature]: Hetzner Console DNS (Beta) - Provide a way to manage the SOA record

I want to manage the whole zone with TF including the SOA record, for example to change the TTL for negative caching (MINIMUM).

That is not possible at the moment as the SERIAL is automatically increased (as it should be), so there will always be a drift.

Maybe I could use the [`hcloud_zone_rrset` data source](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/data-sources/zone_rrset) and pull the current serial, but perhaps it is possible to support this use-case in a nicer way?

jooola:
Because SOA records are created and destroyed with the parent zone, we now:
- import the SOA RRSet in the state, instead of creating it.
- remove the SOA RRSet from the state, instead of deleting it.
- set the SOA record SERIAL value to 0 before saving it to the state.

Closes #1222

### DIFF
--- a/docs/resources/zone_rrset.md
+++ b/docs/resources/zone_rrset.md
@@ -9,7 +9,8 @@ description: |-
   RRSets of type SOA:
   SOA records are created or deleted by the Hetzner Cloud API when creating or deleting
   the parent Zone, therefor this Terraform resource will:
-  import the RRSet in the state, instead of creating it.remove the RRSet from the state, instead of deleting it.set the SOA record SERIAL value to 0 before saving it to the state.
+  import the RRSet in the state, instead of creating it.remove the RRSet from the state, instead of deleting it.set the SOA record SERIAL value to 0 before saving it to the state, as this value is automatically
+  incremented by the API and would cause issues otherwise.
   Experimental: DNS API is in beta, breaking changes may occur within minor releases.
   See https://docs.hetzner.cloud/changelog#2025-10-07-dns-beta for more details.
 ---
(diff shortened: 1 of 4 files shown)


jooola:
Missing:
- ~~Always use 0 as SOA serial value and ensure consistent state after update. Serial format is document in https://docs.hetzner.cloud/reference/cloud#zones section "SOA Serial"~~

docs/resources/zone_rrset.md
```diff
@@ -18,6 +22,15 @@ This can be used to create, modify, and delete Zone RRSets.
...
 See the [Zone RRSets API documentation](https://docs.hetzner.cloud/reference/cloud#zone-rrsets) for more details.
 
+**RRSets of type SOA:**
+
+SOA records are created or deleted by the Hetzner Cloud API when creating or deleting
+the parent Zone, therefor this Terraform resource will:
+
+- import the RRSet in the state, instead of creating it.
+- remove the RRSet from the state, instead of deleting it.
+- set the SOA record SERIAL value to 0 before saving it to the state.
```
apricote: I think this could be explained a bit more thoroughly

```suggestion
- set the SOA record SERIAL value to 0 before saving it to the state, as this value is automatically incremented by the API and would cause issues otherwise.
```

apricote (approved):
🚀

apricote (approved):

PR #1228, opened 2025-10-23
feat: add `zone_record` resource

### LINKED ISSUE #1221: [Feature]: Hetzner Console DNS (Beta) - Ressource record sets should support distributed Terraform states

Dear Hetzner team,

the new Hetzner Cloud DNS implementation works awesome - even if it's still in Beta! Please continue your great work!

There is one little thing that bothers me: The new RRSet logic seems to be incompatible to multiple Terraform states, which are caused by a multi-repo-approach.

**What do I mean exactly?**

I'm managing one repository with IaC code per service. This repos contains the definitions for S3 buckets, compute, DNS and so on for one specific service.

```
Repository "mail-infra" --> TF-state "mail-infra"
|-main.tf
|-dns.tf
|-compute.tf
└-s3.tf

Repository "wiki-infra" --> TF-state "mail-infra"
|-main.tf
|-dns.tf
|-compute.tf
└-s3.tf
```

In this scenario, if both services need a TXT-record `@.mydomain.com`, i.e. `mail` for SPF and `wiki` (SaaS) to verify domain ownership, Terraform isn't able to create the TXT record for `wiki`, because the resource record `TXT` `@.mydomain.com` is managed by the Terraform state of `mail`. The API returns `record not unique` - as expected.

With the old DNS console this wasn't a thing - every record counted as its own resource.

**Which feature(s) I would like to see?**

It would be awesome if the management of resource record sets would be possible with this multi-repo-approach again.

I think of something like this:

**Create**

The provider detects, if the resource record `TXT` `@.mydomain.com` already exists.

_If no:_ Create the resource record and add the value.
_If yes:_ Just adds the additional value.

**Destroy**

The provider detects, if the resource record contains more than one value.

_If yes:_ Just remove the value.
_If no:_ Remove the value and the resource record.

I really appreciate your feedback and your support!

Kind regards
Tom

jooola:
Closes #1221


jooola:
On hold until we decided whether the API receives a new endpoint to support this resource.

hcloud/plugin_provider.go
```diff
@@ -240,6 +241,7 @@ func (p *PluginProvider) Resources(_ context.Context) []func() resource.Resource
 		storageboxsubaccount.NewResource,
 		zone.NewResource,
 		zonerrset.NewResource,
+		zonerecord.NewResource,
```
apricote, line 243: alphabetical sorting (see import)

```suggestion
		zone.NewResource,
		zonerecord.NewResource,
		zonerrset.NewResource,
```

apricote (approved):

Issue #1144, opened 2025-06-27
feat: support for Storage Boxes


kominoshja:
### What whould you like to see?

They were recently migated to the Cloud API, and this would enable provisioning StorageBoxes through TF

apricote:
Hello,

thank you for your interest in the new Storage Box API.

We plan to add support for Storage Boxes to all the official integrations (like this Terraform Provider). As this involves a bit more than just adding new resources, like the new API URL api.hetzner.com, this will take a bit longer than support for other resources. We are currently in the planning phase and can not give you a guaranteed time for when Storage Box support will be available.

Kind Regards,
The Open Source Integrations team at @hetznercloud

kominoshja:
Yep, can imagine, but figured an issue for it wouldn't hurt so that if people search on it, they can come across this.

martipoe:
That is great news, looking forward to this!

apricote:
We have now started working on this.

### Integrating the new API

We want to make this transparent for the average user, so they do not have to think about which API URL has the resources they are interested in.

This means we want to publish the Storage Box resources next to all other existing resources:

```terraform
resource "hcloud_storage_box" "backups" {}
data "hcloud_storage_box_types" "all" {}
```

### Progress

As this is a product that is already launched, we can work on the support in public on GitHub, as we do not want to release something half-baked we will keep a longer lived feature branch `storage-boxes` around until all features are ready: #1166.

- [x] https://github.com/hetznercloud/terraform-provider-hcloud/pull/1167
- [x] Storage Box Types
  - [x] CRUD: https://github.com/hetznercloud/terraform-provider-hcloud/pull/1234
- [x] Storage Boxes
  - [x] CRUD https://github.com/hetznercloud/terraform-provider-hcloud/pull/1250
  - [x] Actions https://github.com/hetznercloud/terraform-provider-hcloud/pull/1251
- [x] Storage Box Subaccounts
  - [x] CRUD https://github.com/hetznercloud/terraform-provider-hcloud/pull/1276
  - [x] Actions https://github.com/hetznercloud/terraform-provider-hcloud/pull/1277
- [x] Storage Box Snapshots
  - [x] CRUD https://github.com/hetznercloud/terraform-provider-hcloud/pull/1272

ToshY:
Hey @apricote 👋 

I'd love to use this. Do you have an estimation when #1276 , #1277 and #1272 are done or when we can expect a new release with this feature?

[...]

PR #1234, opened 2025-10-28
feat(storage-box): implement storage box type datasource

apricote:
Related to #1144


internal/util/experimental/experimental.go
```diff
@@ -64,5 +64,6 @@ func (e Experimental) AppendNotice(description *string) {
 }
 
 var (
-	DNS = New("DNS API", "in beta", "https://docs.hetzner.cloud/changelog#2025-10-07-dns-beta")
+	DNS        = New("DNS API", "in beta", "https://docs.hetzner.cloud/changelog#2025-10-07-dns-beta")
+	StorageBox = New("Storage Box Integration", "experimental", "TODO") // TODO: Create and link ticket before release
```
jooola:
```suggestion
	StorageBox = New("Storage Box implementation", "experimental", "TODO") // TODO: Create and link ticket before release
```

Sounds better. I am not sure if we should stick to the 2 strings (product and maturity), we could use a free text argument, ensuring we cover most if not all cases.

internal/storageboxtype/model.go
```diff
@@ -0,0 +1,108 @@
...
+		var value model
+		diags.Append(value.FromAPI(ctx, item)...)
+
+		tfItem, newDiags := value.ToTerraform(ctx)
+		diags.Append(newDiags...)
+
+		tfItems = append(tfItems, tfItem)
+	}
+
+	m.ID = types.StringValue(datasourceutil.ListID(tfIDs))
```
jooola: Is this really necessary? I understood that we could get ride of those, specially for new resource/datasources?

internal/storageboxtype/model.go
```diff
@@ -0,0 +1,108 @@
...
+	return types.ObjectValueFrom(ctx, m.tfAttributesTypes(), m)
+}
+
+func intToInt64Ptr(i *int) *int64 {
+	if i == nil {
+		return nil
+	}
+	v := int64(*i)
+	return &v
+}
```
jooola, line 76: Why did we need to have different kinds of int between the hcloud-go and terraform?

jooola (approved):

internal/storageboxtype/model.go
apricote: Terraform Plugin Framework only has `int64` and `int32`, while we used an `int` in hcloud-go. This was not noticed during the initial development in hcloud-go as we did not develop the Terraform Provider in parallel.

While we could fix it for Storage Boxes, we will eventually run into this issue while porting other resources to the Plugin Framework.

internal/storageboxtype/model.go
apricote: True, working with terraform to rarely.. Removed the attribute in https://github.com/hetznercloud/terraform-provider-hcloud/pull/1234/commits/5cd507d1765ffde495c407c67f6ef70afb8b2abe

internal/storageboxtype/model.go
jooola: Could we move the function somewhere more relevant to at some point reuse it?

internal/util/experimental/experimental.go
apricote: I will call it `Storage Box support` to match the CLI.

internal/storageboxtype/model.go
apricote: I would keep it here for now, and move it somewhere else if we see that we need it for one of the other storage box resources.

I added tests for the function in a recent commit.

PR #1242, opened 2025-11-04
feat: attach load balancer to a specific network subnet

jooola:
Allows the users to attach a load balancer to a given network subnet.

Using the subnet_id argument is preferred over the network_id argument.


jooola:
Depends on #1246

apricote (approved):

apricote (approved):

PR #1246, opened 2025-11-07
refactor: move resource.hcloud_load_balancer_network to Plugin Framework

jooola:


docs/resources/load_balancer_network.md
```diff
@@ -1,81 +1,66 @@
...
+### Required
+
+- `load_balancer_id` (Number) ID of the Load Balancer.
+
+### Optional
+
+- `enable_public_interface` (Boolean) Wether the Load Balancer public interface is enabled. Default is `true`.
+- `ip` (String) IP to assign to the Load Balancer.
+- `network_id` (Number) ID of the Network to attach the Load Balancer to. Using `subnet_id` is preferred. Required if `subnet_id` is not set. If `subnet_id` or `ip` are not set, the Load Balancer will be attached to the last subnet (ordered by `ip_range`).
+- `subnet_id` (String) ID of the Subnet to attach the Load Balancer to. Required if `network_id` is not set.
```
apricote, line 53: The docs still assume that #1242 is also implemented, but I think that's fine.

docs/resources/load_balancer_network.md
```diff
@@ -1,81 +1,66 @@
...
+
+- `id` (String) The ID of this resource.
 
 ## Import
 
-Load Balancer Network entries can be imported using a compound ID with the following format:
-`<load-balancer-id>-<network-id>`
+Import is supported using the following syntax:
+
+The [`terraform import` command](https://developer.hashicorp.com/terraform/cli/commands/import) can be used, for example:
```
apricote, line 72: Similar lines?

internal/loadbalancer/resource_network.go
```diff
@@ -2,358 +2,531 @@ package loadbalancer
...
+			Validators: []validator.String{
+				validateutil.IP(),
 			},
 		},
+		// XXX: Move to `load_balancer` since it is unrelated to the given private
+		// network attachment.
+		"enable_public_interface": schema.BoolAttribute{
+			MarkdownDescription: "Wether the Load Balancer public interface is enabled. Default is `true`.",
+			Optional:            true,
+			Computed:            true,
```
apricote, line 113: Should not need `computed` with the default, as we always write it -> it always matches the config.

docs/resources/load_balancer_network.md
jooola: This is generated by the terraform docs plugin:

https://github.com/hashicorp/terraform-plugin-docs/blob/main/internal/provider/template.go#L349-L370

We might be missing the tf import block example (`import-by-string-id.tf` I think).

internal/loadbalancer/resource_network.go
jooola: Default does require the computed to be true, or the provide fails with a developer error.

docs/resources/load_balancer_network.md
jooola: https://github.com/hashicorp/terraform-plugin-docs/blob/main/README.md?plain=1#L242-L244

internal/loadbalancer/resource_network.go
jooola:
```
    /home/jo/code/github.com/hetznercloud/terraform-provider-hcloud/internal/loadbalancer/resource_network_test.go:126: Step 1/4 error: Error running pre-apply plan: exit status 1

        Error: Failed to load plugin schemas

        Error while loading schemas for plugin components: Failed to obtain provider
        schema: Could not load the schema for provider
        registry.terraform.io/hashicorp/hcloud: failed to retrieve schema from
        provider "registry.terraform.io/hashicorp/hcloud": Schema Using Attribute
        Default For Non-Computed Attribute: Attribute "enable_public_interface" must
        be computed when using default. This is an issue with the provider and should
        be reported to the provider developers...
        
```

docs/resources/load_balancer_network.md
jooola: I agree.

internal/loadbalancer/resource_network.go
apricote: TIL, thanks.

apricote (approved):

PR #1250, opened 2025-11-11
feat(storage-box): implement storage box resource and datasources

apricote:
Add the new resource `hcloud_storage_box`, `data.hcloud_storage_box` and `data.hcloud_storage_boxes` with the CRUD endpoints:

- https://docs.hetzner.cloud/reference/hetzner#storage-boxes-list-storage-boxes
- https://docs.hetzner.cloud/reference/hetzner#storage-boxes-create-a-storage-box
- https://docs.hetzner.cloud/reference/hetzner#storage-boxes-get-a-storage-box
- https://docs.hetzner.cloud/reference/hetzner#storage-boxes-update-a-storage-box
- https://docs.hetzner.cloud/reference/hetzner#storage-boxes-delete-storage-box

I already started writing tests for things that require the actions to be implemented. They are commented out.

Related to #1144.


apricote:
The [List Folders](https://docs.hetzner.cloud/reference/hetzner#storage-boxes-list-folders-of-a-storage-box) data is not available. I was not sure if it makes to expose in terraform, as it is highly volatile. We could make it an additional data source if someone presents a good use case later on.

internal/storagebox/model.go
```diff
@@ -0,0 +1,221 @@
...
+func (m *modelSnapshotPlan) ToAPI(_ context.Context) (hc *hcloud.StorageBoxSnapshotPlan, diags diag.Diagnostics) {
+	hc = &hcloud.StorageBoxSnapshotPlan{}
+
+	hc.MaxSnapshots = int(m.MaxSnapshots.ValueInt32())
+	hc.Minute = int(m.Minute.ValueInt32())
+	hc.Hour = int(m.Hour.ValueInt32())
+
+	if !m.DayOfWeek.IsNull() {
+		hc.DayOfWeek = hcloud.Ptr(time.Weekday(int(m.DayOfWeek.ValueInt32())))
+	}
```
jooola, line 210: Note to ourselves: Only valid because the API default is null.

internal/storagebox/resource.go
```diff
@@ -0,0 +1,456 @@
...
+			PlanModifiers: []planmodifier.String{
+				stringplanmodifier.RequiresReplace(),
+			},
+		},
+		"password": schema.StringAttribute{
+			MarkdownDescription: "Password of the Storage Box. For more details, see the [Storage Boxes password policy](https://docs.hetzner.cloud/reference/hetzner#storage-boxes-password-policy).",
+			Required:            true,
+			Sensitive:           true,
+			// TODO: Should we generate a password and make it available to the user? Feels painful to always generate a random password, and it can be reset if lost
+		},
```
jooola, line 117: Does it make sense to use a write-only attribute?

internal/storagebox/resource.go
```diff
@@ -0,0 +1,456 @@
...
+			MarkdownDescription: "SSH public keys in OpenSSH format to inject into the Storage Box. It is not possible to update the SSH Keys through the API after creating the Storage Box, so changing this attribute will delete and re-create the Storage Box, you can also add the SSH Keys to the Storage Box manually.",
+			ElementType:         types.StringType,
+			Optional:            true,
+			Computed:            true,
+			Default:             listdefault.StaticValue(types.ListValueMust(types.StringType, []attr.Value{})), // TODO: is there some easier way to get an empty list?
+			PlanModifiers: []planmodifier.List{
+				// No way to update them through the API, similar to servers
+				listplanmodifier.RequiresReplace(),
+			},
+		},
```
jooola: Should we use a `Set` instead?

- No order
- Unique elements

internal/storagebox/resource.go
```diff
@@ -0,0 +1,456 @@
...
+						Older Snapshots will be deleted.
+					`),
+					Required: true,
+				},
+				"minute": schema.Int32Attribute{
+					MarkdownDescription: "Minute when the Snapshot Plan is executed (UTC).",
+					Required:            true,
+				},
+				"hour": schema.Int32Attribute{
+					MarkdownDescription: "Hour when the Snapshot Plan is executed (UTC).\n",
```
jooola:
```suggestion
					MarkdownDescription: "Hour when the Snapshot Plan is executed (UTC).",
```

internal/storagebox/resource.go
```diff
@@ -0,0 +1,456 @@
...
+
+func (m *resourceModel) ToTerraform(ctx context.Context) (types.Object, diag.Diagnostics) {
+	return types.ObjectValueFrom(ctx, m.tfAttributesTypes(), m)
+}
+
+func (r *Resource) ConfigValidators(_ context.Context) []resource.ConfigValidator {
+	return []resource.ConfigValidator{
+		resourcevalidator.Conflicting(
+			path.MatchRoot("snapshot_plan").AtName("day_of_week"),
+			path.MatchRoot("snapshot_plan").AtName("day_of_month"),
```
jooola: Are you certain about this? I think one can define both day of week and day of month?

The API allows it at least, even if it might not make 100% sens.

internal/storagebox/resource.go
```diff
@@ -0,0 +1,456 @@
...
+		"ssh_keys": schema.ListAttribute{
+			MarkdownDescription: "SSH public keys in OpenSSH format to inject into the Storage Box. It is not possible to update the SSH Keys through the API after creating the Storage Box, so changing this attribute will delete and re-create the Storage Box, you can also add the SSH Keys to the Storage Box manually.",
+			ElementType:         types.StringType,
+			Optional:            true,
+			Computed:            true,
+			Default:             listdefault.StaticValue(types.ListValueMust(types.StringType, []attr.Value{})), // TODO: is there some easier way to get an empty list?
+			PlanModifiers: []planmodifier.List{
+				// No way to update them through the API, similar to servers
+				listplanmodifier.RequiresReplace(),
+			},
```
jooola: No preference in my head, but would like to discuss this:

I am tempted to go the "ignore ssh keys except for creation" route, because deleting a storage box is a rare occurrence or the user might loose data. Forcing the user to maintain the ssh public keys after the resource was created, seem more annoying to me.

internal/storagebox/resource.go
```diff
@@ -0,0 +1,456 @@
...
+				"hour": schema.Int32Attribute{
+					MarkdownDescription: "Hour when the Snapshot Plan is executed (UTC).\n",
+					Required:            true,
+				},
+				"day_of_week": schema.Int32Attribute{
+					// TODO: Also accept string days similar to CLI?
+					MarkdownDescription: util.MarkdownDescription(`
+						Day of the week when the Snapshot Plan is executed.
+
+						Starts at 1 for Monday til 7 for Sunday. Null means every day.
```
jooola: We are mapping the go's week day to the API week day in hcloud-go, so this is not correct.

docs/data-sources/storage_box.md
```diff
@@ -0,0 +1,80 @@
...
+- `username` (String) Primary username of the Storage Box.
+
+<a id="nestedatt--snapshot_plan"></a>
+### Nested Schema for `snapshot_plan`
+
+Read-Only:
+
+- `day_of_month` (Number) Day of the month when the Snapshot Plan is executed.
+
+Null means every day.
```
jooola: Should we remove the empty line here? I think the new line is good, but the extra new lines might be too much.

jooola (commented):
First pass on the whole thing, looks good ;)

A few remarks though.

internal/storagebox/resource.go
apricote: I thought about it, but how could we figure out that it changed and that we need to call the reset password action?

internal/storagebox/resource.go
apricote: True, I somehow had it in my mind that it was not possible to specify both. Will remove the validation.

docs/data-sources/storage_box.md
apricote: I think the newlines here make it less readable, as the are missing the indent and break the list. I will remove them

internal/storagebox/model.go
apricote: can you explain this comment?

internal/storagebox/resource.go
jooola: Yeah true, didn't look much into it, but it is indeed not possible.

https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments#general-concepts

internal/storagebox/resource.go
apricote: Good idea :) I was thinking too hard about the API schema.

docs/data-sources/storage_box.md
jooola: Could also be fixed on the terraform plugin docs side.

internal/storagebox/resource.go
apricote: Updated the comment and added a hint about the divergence from the API.

internal/storagebox/model.go
jooola: When writing to the API, the above code should support send either null, or the int value. The current code works only because null is also the default value.

Just wanted to point it out, as I think this is the first time we have such a use case. Also I think hcloud-go is handling this for us, and always send nil, when unset.

internal/storagebox/resource.go
apricote: And return a warning diagnostic if there is a diff? Because if the user explicitly tries to change the SSH Keys, but the changes are just never applied then that's also a bad experience.

internal/storagebox/resource.go
```diff
@@ -0,0 +1,433 @@
...
+		return
+	}
+
+	// Create in API
+	result, _, err := r.client.StorageBox.Create(ctx, opts)
+	if err != nil {
+		resp.Diagnostics.Append(hcloudutil.APIErrorDiagnostics(err)...)
+		return
+	}
+
```
jooola, line 320: Consider saving the ID to the state.

Following https://github.com/hetznercloud/terraform-provider-hcloud/pull/1257

internal/storagebox/model.go
apricote: Alright, thanks for explaining :)

internal/storagebox/resource.go
apricote: Will be in the next push.

internal/storagebox/resource.go
apricote: In the latest push I changed the behaviour to ignore/hide the diff when the user changes the attribute and instead "just" show a warning diagnostic.

docs/data-sources/storage_box.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Optional
+
+- `id` (Number) ID of the Storage Box.
+- `name` (String) Name of the Storage Box.
+- `snapshot_plan` (Attributes) Details of the active snapshot plan. (see [below for nested schema](#nestedatt--snapshot_plan))
```
jooola: This should be read only, something in the schema is invalid.

docs/data-sources/storage_boxes.md
```diff
@@ -0,0 +1,79 @@
...
+### Read-Only
+
+- `storage_boxes` (Attributes List) (see [below for nested schema](#nestedatt--storage_boxes))
+
+<a id="nestedatt--storage_boxes"></a>
+### Nested Schema for `storage_boxes`
+
+Optional:
+
+- `snapshot_plan` (Attributes) Details of the active snapshot plan. (see [below for nested schema](#nestedatt--storage_boxes--snapshot_plan))
```
jooola: This should be read only, something in the schema is invalid.

internal/storagebox/data_source.go
```diff
@@ -0,0 +1,255 @@
...
+			Computed:            true,
+		},
+		"delete_protection": schema.BoolAttribute{
+			MarkdownDescription: "Whether delete protection is enabled.",
+			Computed:            true,
+		},
+		"snapshot_plan": schema.SingleNestedAttribute{
+			MarkdownDescription: "Details of the active snapshot plan.",
+			Optional:            true,
+			Computed:            true,
```
jooola: This should be read only.

internal/storagebox/data_source.go
```diff
@@ -0,0 +1,255 @@
...
+			resp.Diagnostics.Append(hcloudutil.APIErrorDiagnostics(err)...)
+			return
+		}
+		if result == nil {
+			resp.Diagnostics.Append(hcloudutil.NotFoundDiagnostic("storage box", "name", data.Name.String()))
+			return
+		}
+	case !data.WithSelector.IsNull():
+		opts := hcloud.StorageBoxListOpts{}
+		opts.LabelSelector = data.WithSelector.ValueString()
```
jooola, line 237: I understand that if unset value string is the go empty value and we don't send it. But reading this is not obvious that it is optional, and is an bit more involved while reviewing than having a condition around it.

Happy to keep like this, just wanted to point that out.

internal/storagebox/data_source_list.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// Schema should return the schema for this data source.
+func (d *DataSourceList) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema.MarkdownDescription = `
+Provides a list of Hetzner Storage Boxes.
+
+See the [Storage Boxes API documentation](https://docs.hetzner.cloud/reference/hetzner#storage-boxes) for more details.
+`
+
```
jooola, line 59: I think we are missing the experimental marker here.

internal/storagebox/data_source.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+// Schema should return the schema for this data source.
+func (d *DataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema.MarkdownDescription = `
+Provides details about a Hetzner Storage Box.
+
+See the [Storage Boxes API documentation](https://docs.hetzner.cloud/reference/hetzner#storage-boxes) for more details.
+`
+
```
jooola, line 161: I think we are missing the experimental marker here.

internal/storagebox/resource.go
```diff
@@ -0,0 +1,471 @@
...
+			},
+		},
+		"delete_protection": schema.BoolAttribute{
+			MarkdownDescription: "Prevent the Storage Box from being accidentally deleted outside of Terraform.",
+			Optional:            true,
+			Computed:            true,
+			Default:             booldefault.StaticBool(false),
+		},
+		"snapshot_plan": schema.SingleNestedAttribute{
+			MarkdownDescription: "",
```
jooola: Should we keep this if its empty?

jooola (commented):
I noticed we don't implement the resource identity here, so I assume the record resource was only a one off or should we push for having identities everywhere?

jooola:
Aside from the little comments, it looks really good :rocket:

apricote:
> I noticed we don't implement the resource identity here, so I assume the record resource was only a one off or should we push for having identities everywhere?

We noticed that the Identity does not work at all with OpenTofu. IIRC we decided to significantly scale down the usage of it to a level where its more annoying to have/use then anything else. As we can do a simple import with the ID here, I did not see a necessity to add the identity.

internal/storagebox/data_source.go
apricote: Check. I forgot that `Computed: true` still means that it can be null, and added to `Optional: true` to indicate that it may be null.

internal/storagebox/resource.go
apricote: Nope, missed to add the actual description. Will be fixed in the next push.

apricote: Fixed in [609a8ff](https://github.com/hetznercloud/terraform-provider-hcloud/pull/1250/commits/609a8ffb93b589c4f7b80c1441b6619566600376)

internal/storagebox/data_source.go
apricote: Fixed in [609a8ff](https://github.com/hetznercloud/terraform-provider-hcloud/pull/1250/commits/609a8ffb93b589c4f7b80c1441b6619566600376)

internal/storagebox/data_source_list.go
apricote: Fixed in [609a8ff](https://github.com/hetznercloud/terraform-provider-hcloud/pull/1250/commits/609a8ffb93b589c4f7b80c1441b6619566600376)

internal/storagebox/data_source.go
apricote: Fixed in [609a8ff](https://github.com/hetznercloud/terraform-provider-hcloud/pull/1250/commits/609a8ffb93b589c4f7b80c1441b6619566600376)

docs/data-sources/storage_boxes.md
apricote: Fixed in [609a8ff](https://github.com/hetznercloud/terraform-provider-hcloud/pull/1250/commits/609a8ffb93b589c4f7b80c1441b6619566600376)

docs/data-sources/storage_box.md
apricote: Fixed in [609a8ff](https://github.com/hetznercloud/terraform-provider-hcloud/pull/1250/commits/609a8ffb93b589c4f7b80c1441b6619566600376)

PR #1251, opened 2025-11-11
feat(storage-box): add actions to hcloud_storage_box resource

apricote:
Add the actions for `storage_boxes` to the `hcloud_storage_box` resource.

- https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-change-protection
- https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-change-type
- https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-reset-password
- https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-update-access-settings
- https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-disable-snapshot-plan
- https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-enable-snapshot-plan

Related to #1144.

Notably missing is [Rollback Snapshot](https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-rollback-snapshot). I was not sure how to best implement this "imperative" one-off action. There is no way for us to check if it needs to be done, and snapshots are automatically deleted after a while, making it unsuitable for terraform. [Terraform `actions`](https://developer.hashicorp.com/terraform/language/v1.14.x/invoke-actions?product_intent=terraform) might be a better fit for this, but they are still in beta in Terraform and [not supported by OpenTofu](https://github.com/opentofu/opentofu/issues/3309).

### DIFF
--- a/internal/storagebox/resource.go
+++ b/internal/storagebox/resource.go
@@ -323,14 +323,53 @@ func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp 
 	// surprise "duplicate resource" errors happen.
 	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), types.Int64Value(result.StorageBox.ID))...)
 
-	if err = r.client.Action.WaitFor(ctx, result.Action); err != nil {
-		resp.Diagnostics.Append(hcloudutil.APIErrorDiagnostics(err)...)
+	resp.Diagnostics.Append(hcloudutil.SettleActions(ctx, &r.client.Action, result.Action)...)
+	if resp.Diagnostics.HasError() {
 		return
 	}
 
-	// TODO: (for Actions PR)
-	//   Set Delete Protection
-	//   Enable Snapshot Plan
+	var actions []*hcloud.Action
+
+	if !data.DeleteProtection.IsUnknown() && !data.DeleteProtection.IsNull() {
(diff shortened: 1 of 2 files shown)


internal/storagebox/resource.go
```diff
@@ -323,14 +323,53 @@ func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp
 	// surprise "duplicate resource" errors happen.
 	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), types.Int64Value(result.StorageBox.ID))...)
 
-	if err = r.client.Action.WaitFor(ctx, result.Action); err != nil {
-		resp.Diagnostics.Append(hcloudutil.APIErrorDiagnostics(err)...)
+	resp.Diagnostics.Append(hcloudutil.SettleActions(ctx, &r.client.Action, result.Action)...)
+	if resp.Diagnostics.HasError() {
 		return
 	}
```
jooola, line 329: Can we not wait here? Once the storage box is created we might be able to hit it with more actions straight away?

jooola (approved):
🪨

internal/storagebox/resource.go
apricote: 2 out of 2 e2e test runs for this failed in step 5, where we test resource creation + immediately running actions. I will keep it as is and wait for the create action before doing anything else as that has been stable.

PR #1252, opened 2025-11-11
refactor: replace hcloudutil.WaitForAction(s)

apricote:
These two functions used to be more complex, but since #926 they are just a wrapper around hcloud-go that no longer serves any purpose.


jooola:
I thought we wanted to keep the helper and extends it to wait for all actions to complete and add diagnostics for each errors?

apricote:
> I thought we wanted to keep the helper and extends it to wait for all actions to complete and add diagnostics for each errors?

That's a possibility for later. [Some places already have a pattern like this implemented](https://github.com/hetznercloud/terraform-provider-hcloud/pull/1252/files#diff-7378c7d462221547688e0d724658c2450b4489a0b02abcd83ad57b30ffa530c4R454). But most of the old resources just run the actions sequentially, always waiting for it to settle before starting the next. Only the last few resources run the actions in parallel anyway.

internal/loadbalancer/resource_target.go
```diff
@@ -104,19 +104,19 @@ func resourceLoadBalancerTargetCreate(ctx context.Context, d *schema.ResourceDat
...
 	if err != nil {
 		return hcloudutil.ErrorToDiag(err)
 	}
-	if a != nil {
-		if err := hcloudutil.WaitForAction(ctx, &c.Action, a); err != nil {
+	if action != nil {
+		if err = c.Action.WaitFor(ctx, action); err != nil {
 			return diag.Errorf("add load balancer target: %v", err)
 		}
 	}
```
jooola: `if action != nil {` can be skipped, the WaitFor function filters out nil actions.

https://github.com/hetznercloud/hcloud-go/blob/main/hcloud/action_waiter.go#L25

internal/placementgroup/resource.go
```diff
@@ -86,7 +86,7 @@ func resourcePlacementGroupCreate(ctx context.Context, d *schema.ResourceData, m
 	d.SetId(util.FormatID(res.PlacementGroup.ID))
 
 	if res.Action != nil {
-		if err := hcloudutil.WaitForAction(ctx, &client.Action, res.Action); err != nil {
+		if err = client.Action.WaitFor(ctx, res.Action); err != nil {
 			return hcloudutil.ErrorToDiag(err)
 		}
```
jooola: Skip nil check.

internal/primaryip/resource.go
```diff
@@ -126,7 +126,7 @@ func resourcePrimaryIPCreate(ctx context.Context, d *schema.ResourceData, m inte
 
 	d.SetId(util.FormatID(res.PrimaryIP.ID))
 	if res.Action != nil {
-		if err := hcloudutil.WaitForAction(ctx, &client.Action, res.Action); err != nil {
+		if err := client.Action.WaitFor(ctx, res.Action); err != nil {
 			return hcloudutil.ErrorToDiag(err)
 		}
```
jooola: Skip nil check.

internal/server/resource.go
```diff
@@ -447,12 +447,12 @@ func resourceServerCreate(ctx context.Context, d *schema.ResourceData, m interfa
...
 	}
 
-	if err := hcloudutil.WaitForAction(ctx, &c.Action, res.Action); err != nil {
+	if err = c.Action.WaitFor(ctx, res.Action); err != nil {
 		diags = append(diags, hcloudutil.ErrorToDiag(err)...)
 		return
 	}
 	for _, nextAction := range res.NextActions {
-		if err := hcloudutil.WaitForAction(ctx, &c.Action, nextAction); err != nil {
+		if err = c.Action.WaitFor(ctx, nextAction); err != nil {
```
jooola, line 455: You may use https://pkg.go.dev/github.com/hetznercloud/hcloud-go/v2@v2.30.0/hcloud/exp/actionutil#AppendNext to merge the multiple actions together.

jooola (commented):
🧹

jooola (approved):

internal/loadbalancer/resource_target.go
apricote: nice, removed all nil checks.

internal/server/resource.go
apricote: I left this function the same on purpose, as it is a one-off implementation of "settle", where it collects the results of each action individually. Moving to a single `c.Action.WaitFor(ctx, res.NextActions)` would make the error messages worse.

jooola (approved):

PR #1268, opened 2025-11-19
refactor: add function to wait for all actions to settle

apricote:
Adds a new `hcloudutil.SettleActions()` helper that uses `ActionClient.WaitForFunc` to wait until all actions have failed or completed.

Any actions that have failed are then presented as nice error diagnostics, similar to how we already do it with synchronous API errors.

If the error returned from `ActionClient.WaitForFunc` matches a cancelled context, we log the actions that we are still waiting for.

I refactored all framework resources to use this. As its is using the framework `diag.Diagnostics` type, it is not easily compatible with SDKv2 resources.

### DIFF
--- a/internal/util/hcloudutil/actions.go
+++ b/internal/util/hcloudutil/actions.go
@@ -2,7 +2,9 @@ package hcloudutil
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-framework/diag"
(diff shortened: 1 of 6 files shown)


internal/util/hcloudutil/actions.go
```diff
@@ -0,0 +1,58 @@
...
+		switch update.Status {
+		case hcloud.ActionStatusError:
+			failedActions = append(failedActions, update)
+		}
+
+		return nil
+	}, actions...)
+	if err != nil {
+		return APIErrorDiagnostics(err)
+	}
```
jooola: Should we catch a potential context deadline / cancelled to pretty print actions that "timed out" (client side)?

apricote: Sure, will add that.

apricote: Added in 144078fe89491c0596358074322c9f3e27ec46a6

jooola (approved):
🚀

jooola:
image
Nice

PR #1269, opened 2025-11-19
ci: update support terraform version matrix

jooola:
```rp-commits
feat: drop support for terraform v1.11
feat: drop support for terraform v1.12
feat: add support for terraform v1.13
feat: add support for terraform v1.14
```


apricote (approved):
Testing against 6 versions 💀

jooola:
> Testing against 6 versions 💀

I am not sure anymore, according to TF, they only support 2 major version so maybe we should drop 1.12 too.

PR #1272, opened 2025-11-20
feat(storage-box): implement snapshot resource and datasources

apricote:
Add the new resources `hcloud_storage_box_snapshot`, `data.hcloud_storage_box_snapshot` and `data.hcloud_storage_box_snapshots` with the CRUD endpoints:

- https://docs.hetzner.cloud/reference/hetzner#storage-box-snapshots-list-snapshots
- https://docs.hetzner.cloud/reference/hetzner#storage-box-snapshots-create-a-snapshot
- https://docs.hetzner.cloud/reference/hetzner#storage-box-snapshots-create-a-snapshot
- https://docs.hetzner.cloud/reference/hetzner#storage-box-snapshots-update-a-snapshot
- https://docs.hetzner.cloud/reference/hetzner#storage-box-snapshots-delete-a-snapshot

Related to #1144.


internal/storageboxsnapshot/resource.go
```diff
@@ -0,0 +1,286 @@
...
+
+	resp.Schema.Attributes = map[string]schema.Attribute{
+		"storage_box": schema.Int64Attribute{
+			MarkdownDescription: "ID of the Storage Box.",
+			Required:            true,
+			PlanModifiers: []planmodifier.Int64{
+				int64planmodifier.UseStateForUnknown(),
+				int64planmodifier.RequiresReplace(),
+			},
+		},
```
jooola: About a `_id` suffix.

In other resource e.g. in `server_network` we use `server_id` and `network_id`. 

For server, we use `server_type`, because we should accept ID or Name, but it seem we only accept the name.

What is the pattern that we should follow here? I'd rather only accept an ID, so we don't have to fetch the API ourselves for each snapshot resource. But in that case maybe a id suffix could be better?

internal/storagebox/resource_test.go
```diff
@@ -21,42 +20,21 @@ import (
...
-
 func TestAccStorageBoxResource(t *testing.T) {
 	tmplMan := testtemplate.Manager{}
 
 	storageBox := &hcloud.StorageBox{}
 
 	res := &storagebox.RData{
 		StorageBox: schema.StorageBox{
-			Name:           fmt.Sprintf("storage-box-%s", randutil.GenerateID()),
+			Name:           fmt.Sprintf("snapshot-%s", randutil.GenerateID()),
```
jooola: Upsy?

jooola (approved):

internal/storageboxsnapshot/resource.go
apricote: Ack, I think I only looked at `server`, but nearly all other references (to user-resources) use `_id`.

internal/storagebox/resource_test.go
apricote: Upsy indeed.

apricote: Resolved in 0aedaa6e35d8d4d3f783bc81ccc43e102eca5e24

internal/storageboxsnapshot/resource.go
apricote: Resolved in 0aedaa6e35d8d4d3f783bc81ccc43e102eca5e24

PR #1273, opened 2025-11-20
test: add testtemplate.DeepCopy helper

jooola:
This can be useful to copy all fields of a test resource and only override some fields for a next test step.

### DIFF
--- a/internal/server/resource_test.go
+++ b/internal/server/resource_test.go
@@ -265,66 +265,57 @@ func TestAccServerResource_Resize(t *testing.T) {
 func TestAccServerResource_ChangeUserData(t *testing.T) {
 	tmplMan := testtemplate.Manager{}
 
-	var s, s2 hcloud.Server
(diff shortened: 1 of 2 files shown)


internal/server/resource_test.go
```diff
@@ -1401,15 +1392,6 @@ func TestAccServerResource_EmptySSHKey(t *testing.T) {
...
 }
 
-func isRecreated(newServer, oldServer *hcloud.Server) func() error {
-	return func() error {
-		if newServer.ID == oldServer.ID {
-			return fmt.Errorf("new server is the same as server cert %d", oldServer.ID)
-		}
-		return nil
-	}
-}
```
jooola, line 1411: Replacement is now checked with:
```
plancheck.ExpectResourceAction(res2.TFID(), plancheck.ResourceActionReplace),
```

apricote (approved):

PR #1275, opened 2025-11-21
style: do not print 4 new lines when action error message is empty

jooola:


internal/util/hcloudutil/actions.go
```diff
@@ -58,15 +58,15 @@ func SettleActions(ctx context.Context, client ActionWaiter, actions ...*hcloud.
...
-			"Resources: %s\n",
-		action.ErrorMessage, action.ErrorCode, action.Command, action.ID, actionResourceDescription(action),
-	))
+	detail := strings.Builder{}
+
+	detail.WriteString("An API action for the resource failed.\n\n")
+	if action.ErrorMessage != "" {
+		detail.WriteString(action.ErrorMessage + "\n\n")
+	}
+	detail.WriteString(actionDetail(action))
```
apricote, line 67: I would personally not make `actionDetail` its own function, and do it inline here instead. Not a blocker, just my opinion.

apricote (approved):

internal/util/hcloudutil/actions.go
jooola: What about reusing it to split the single timeout diagnostic for X actions into multiple ones, and reuse this detail function to have a consistent detail printing when an action failed?

apricote: To me it feels like every single diagnostic has a huge cognitive overhead in Terraform, where Terraform adds 5+ additional lines with more context.

As this context is the same for all timeouted actions, I feel like splitting them up actually makes them less visible, and customers are more likely to not read the wall of text closely.

PR #1276, opened 2025-11-21
feat(storage-box): implement subaccount resource and data sources

apricote:
Add the new resources `hcloud_storage_box_subaccount`, `data.hcloud_storage_box_subaccount` and
`data.hcloud_storage_box_subaccounts` with the CRUD endpoints:

- https://docs.hetzner.cloud/reference/hetzner#storage-box-subaccounts-list-subaccounts
- https://docs.hetzner.cloud/reference/hetzner#storage-box-subaccounts-create-a-subaccount
- https://docs.hetzner.cloud/reference/hetzner#storage-box-subaccounts-get-a-subaccount
- https://docs.hetzner.cloud/reference/hetzner#storage-box-subaccounts-update-a-subaccount
- https://docs.hetzner.cloud/reference/hetzner#storage-box-subaccounts-delete-a-subaccount

Related to #1144.


examples/data-sources/hcloud_storage_box_subaccount/data-source.tf
```diff
@@ -0,0 +1,16 @@
+variable "storage_box_id" {}
+
+data "hcloud_storage_box_subaccount" "by_id" {
+  storage_box = var.storage_box_id
+  id          = 2
+}
+
+data "hcloud_storage_box_subaccount" "by_username" {
+  storage_box = var.storage_box_id
+  username    = "2025-02-12T11-35-19"
```
jooola: This example value is probably confusing.

internal/storageboxsubaccount/data_source.go
```diff
@@ -0,0 +1,217 @@
...
+					MarkdownDescription: "Whether the Subaccount is read-only.",
+					Computed:            true,
+				},
+			},
+		},
+		"labels": schema.MapAttribute{
+			MarkdownDescription: "User-defined [labels](https://docs.hetzner.cloud/reference/cloud#labels) (key-value pairs) for the resource.",
+			ElementType:         types.StringType,
+			Computed:            true,
+		},
```
jooola: Didn't we have a helper to define this?

internal/storageboxsubaccount/data_source_list_test.go
```diff
@@ -0,0 +1,127 @@
...
+		Raw: fmt.Sprintf(`
+			access_settings = {
+				reachable_externally = false
+				samba_enabled = true
+				ssh_enabled = false
+				webdav_enabled = true
+				readonly = false
+			}
+
+			depends_on = [%s]`, res1.TFID()), // Only one subaccount may be created at the same time
```
jooola, line 72: For snapshot I would understand, but for subaccounts, I think we should handle the error in the provider.

Or is this a left over from the copy pasta from snapshots?

internal/storageboxsubaccount/resource.go
```diff
@@ -0,0 +1,387 @@
...
+	result, _, err := r.client.StorageBox.CreateSubaccount(ctx, storageBox, opts)
+	if err != nil {
+		resp.Diagnostics.Append(hcloudutil.APIErrorDiagnostics(err)...)
+		return
+	}
+
+	// Make sure to save the ID immediately so we can recover if the process stops after
+	// this call. Terraform marks the resource as "tainted", so it can be deleted and no
+	// surprise "duplicate resource" errors happen.
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), types.Int64Value(result.Subaccount.ID))...)
```
jooola, line 240: Should we save the parent storage box ID too? e.g. the user removes the resource from their config?

jooola (approved):

internal/storageboxsubaccount/data_source_list_test.go
apricote: Nope, I tested it without the `depends_on`, but it broke in the same way. If its fine for you, I will add it in a follow up PR for both resources.

internal/storageboxsubaccount/data_source.go
apricote: Only for resources :( Not for the `datasource/schema` types.

examples/data-sources/hcloud_storage_box_subaccount/data-source.tf
apricote: Agreed, seems I copied that from the snapshot examples (where it matches the actual snapshot names).

internal/storageboxsubaccount/data_source_list_test.go
jooola: Sure

internal/storageboxsubaccount/data_source.go
apricote: https://github.com/hetznercloud/terraform-provider-hcloud/pull/1281

internal/storageboxsubaccount/resource.go
apricote: Good catch, I assumed terraform would automagically add it to the state, as its a required attribute, but in a test the state only contained was `storage_box_id = null`, which caused the resource to be deleted from state as it could not be read, so it leaked.

Will add the attribute here and check the other resources for the same bug.

apricote:
Failed e2e tests are because of missing server capacity, the storage box tests were successfull.

Issue #1224, opened 2025-10-21
[Feature]: hcloud server (start_after_create)


cloudfy:
### What whould you like to see?

When creating a hcloud server, ability to se start_after_create true/false

jansedlon:
Second this. Right now the servers are created turned off

jooola:
> Second this. Right now the servers are created turned off

This should not be the case, you might be running into a bug. Please open a ticket using the [Support Center](https://www.hetzner.com/support-center).

apricote:
Hey @cloudfy and @jansedlon,

can you explain your use-case for creating turned-off servers in Terraform? This feels at-odds with the declarative approach terraform enforces, so I am curious to hear how you would like to use this.

cloudfy:
Ensuring subnet allocation without having to specify fixed IP address. The way Hetzner made the allocation of network, I cannot assign subnet to a server, ensuring the proper IP when it boots. The use-case is to create server stopped, remove from network, assign to subnet and then start, which boots the server with the right ip.

apricote:
The API now supports attaching directly to a subnet. We added support for it in https://github.com/hetznercloud/terraform-provider-hcloud/pull/1217 for the `hcloud_server_network` resource, but it is missing in `hcloud_server`. Would that solve your requirement?

Is it necessary that the server only boots when it's assigned to the right private network + subnet?

cloudfy:
> Is it necessary that the server only boots when it's assigned to the right private network + subnet?
Proper cloud_init in scalable designs require the subnet to be correct.

I echo the #1217 , "Using the subnet_id argument is preferred over the network_id argument."

Small deployment is fair to use a small network_id, but when it comes to scalable subnets, the subnet is most important. In such case, assigning the subnet to a "stopped" server can be voided, rendering the parameter suggestion (start_after_create) obsolute.

mlinares1998:
> The API now supports attaching directly to a subnet. We added support for it in [#1217](https://github.com/hetznercloud/terraform-provider-hcloud/pull/1217) for the `hcloud_server_network` resource, but it is missing in `hcloud_server`. Would that solve your requirement?
> 
> Is it necessary that the server only boots when it's assigned to the right private network + subnet?

Hi @apricote, @jooola, @cloudfy 

I've opened #1265, which seems related to this issue as well.

[...]

PR #1278, opened 2025-11-24
feat(server): add subnet_id support

### LINKED ISSUE #1265: [Bug]: Cannot assign a private server to a subnet without other IP/Interface assigned
### What happened?

Hi Team,
I've encountered an issue when trying to use the new subnet assignment feature from v1.56.0 (#1217).

When creating a VM without a network block or public IP enabled (expecting hcloud_server_network to handle the subnet assignment), the hcloud_server resource fails. The creation fails as the instance cannot start because it has no interface assigned at creation time.

The VM should be created and then have its network/subnet assigned via hcloud_server_network.

Has anyone else encountered this?
Regards!

Log: 

```bash
 Error: hcloudutil/WaitForAction: no public or private network interfaces found, cannot start server (invalid_input, 593478587344330)
  │ 
  │   with hcloud_server.test,
  │   on test_vm.tf line 1, in resource "hcloud_server" "test":
  │    1: resource "hcloud_server" "test" {
  │ 
```

The VM should be created and then have its network/subnet assigned via hcloud_server_network.

Problematic code
```HCL
resource "hcloud_server" "test" {
  depends_on = [hcloud_network_subnet.subnet]

  name                     = "debian-test-fsn1-staging"
  server_type              = "cax11"
  datacenter               = "fsn1-dc14"
  image                    = "debian-13"
  keep_disk                = true
  backups                  = false
  labels                   = merge(var.labels, {})
  shutdown_before_deletion = true

  ssh_keys = [hcloud_ssh_key.env.id]

  public_net {
    ipv4_enabled = false
    ipv6_enabled = false
  }

  user_data = <<EOF
#cloud-config
write_files:
  - path: /etc/network/interfaces
    content: |
      auto enp7s0
      iface enp7s0 inet dhcp
          post-up echo "Waiting..."
          post-up ip route add default via 10.11.0.1
          dns-nameservers ${var.nat_ip}
    append: true
runcmd:
  - reboot
EOF
}

resource "hcloud_server_network" "node1_subnet1" {
  server_id = hcloud_server.test.id
  subnet_id = "11180245-10.11.95.0/25"
  alias_ips = []
}
```

Enabling public IP fixes the problem.

mlinares1998:
### Summary

This PR adds `subnet_id` support to the `hcloud_server` resource's inline `network` blocks, allowing users to specify which subnet within a network to attach the server to. This brings feature parity with the standalone `hcloud_server_network` resource (PR #1217).

Fixes #1265, #1224


docs/resources/server.md
```diff
@@ -94,6 +94,27 @@ resource "hcloud_server" "server" {
...
+    network_id = hcloud_network.network.id
+    subnet_id  = hcloud_network_subnet.network-subnet.id
+    alias_ips  = []
+  }
+
+  depends_on = [
+    hcloud_network_subnet.network-subnet
+  ]
+}
+```
```
jooola: Could you please make this example the recommended one? Bring it on the top and remove the `depends_one` attribute.

Also we probably want to require oneOf network ID or subnet ID, and not allow both.

docs/resources/server.md
```diff
@@ -228,7 +250,8 @@ The following attributes are exported:
 
 a single entry in `network` support the following fields:
 
-- `network_id` - (Required, int) ID of the network
+- `network_id` - (Required, int) ID of the Network to attach the Server to. If `subnet_id` or `ip` are not set, the Server will be attached to the last subnet (ordered by `ip_range`).
+- `subnet_id` - (Optional, string) ID of the Subnet to attach the Server to. When specified, the server will be attached to this specific subnet within the network. The subnet must belong to the network specified in `network_id`.
```
jooola: Please require one of networkID or subnetID, since the subnet ID is composed of the network ID and the subnet ip range.

Only having a dependency on the subnet ID makes it simpler for the users.

jooola (commented):
Hey, thanks for the fix.

docs/resources/server.md
mlinares1998: Hi! @jooola 
Sounds great!
I have updated the PR implementing this.

mlinares1998: Hi @jooola 
I've updated the documentation to make the subnet_id approach the primary example, with depends_on removed as you suggested. The previous approach (network_id without subnet_id) is now documented as "not recommended" as reference.

mlinares1998 (commented):
Hi @jooola I have implemented the requested changes, could you review the PR again? 😄 
Regards!

mlinares1998:
Hi @jooola,
I’ve merged the latest changes from main.
Let me know if you have any further suggestions or fixes. :smile: 

Thanks!

mlinares1998:
Hi @jooola,

I’m following up on this PR as it’s been a few months since the last review.

This change is a critical dependency for hcloud-k8s/terraform-hcloud-kubernetes#163. Currently, the lack of subnet_id support forces us to hardcode node pool IP addresses. This lacks the necessary flexibility and, more importantly, causes unreliable behavior during node pool deletions.

We can't safely resolve these lifecycle issues in the Talos module without this provider-level support.

I'm happy to make any further adjustments if needed. 
Thanks!

M4t7e:
Hey @jooola, @apricote,

sorry for the noise, but do you think this can be finished within Q1? We currently have a few issues to solve, and we need this feature in order to change the subnetting layout in [hcloud-k8s/terraform-hcloud-kubernetes](https://github.com/hcloud-k8s/terraform-hcloud-kubernetes), which is a prerequisite for solving the issues properly. If it can't, we'll need to replan and likely postpone everything to the following major release (v5), since changing the subnetting layout will be a breaking change that requires user action and we have to release the next major version (v4) for other reasons latest end of Q1.

I don't want to bug you, but I'd appreciate an indication of whether it's worth waiting for this or not :slightly_smiling_face:

marcwickenden:
Hi @jooola and @apricote - any update on this? I'm waiting on this fix so terraform-hcloud-kubernetes can be updated and I can migrate a critical workload onto Hetzner.

jooola:
Hey all,
We are currently really busy with other topics, sorry about that.

I'll try to look at this tomorrow. I can't promise anything though.

laurentpellegrino:
@jooola Kindly ping :heart:

mlinares1998:
> Hey all, We are currently really busy with other topics, sorry about that.
> 
> I'll try to look at this tomorrow. I can't promise anything though.

Hi! @jooola, i have seen you made changes to the branch.
Anything else we could help you to have this merged soon?
Kindly ping about this :heart:

internal/server/resource.go
```diff
@@ -1443,11 +1513,29 @@ func validateUniqueNetworkIDs(d *schema.ResourceDiff) error {
 				return fmt.Errorf("network item has unexpected type: %T", networkI)
 			}
 
-			networkIDI, ok := network["network_id"]
-			if !ok {
-				continue
+			networkID := util.CastInt64(network["network_id"])
```
jooola: Could we be a bit more defensive here?

Before `network_id` was required within the `network` block, but not anymore.

I haven't check if using `util.CastInt64` against an undefined network id is safe.

internal/server/resource.go
```diff
@@ -1275,13 +1295,41 @@ func getServerAttributes(d *schema.ResourceData, s *hcloud.Server, forceSetNetwo
...
 		tfPrivateNetwork := make(map[string]interface{})
-		tfPrivateNetwork["network_id"] = privateNetwork.Network.ID
 		tfPrivateNetwork["ip"] = privateNetwork.IP.String()
 		tfPrivateNetwork["mac_address"] = privateNetwork.MACAddress
+
+		// Check the user input to preserve the same structure in state
+		if nwSet, ok := d.GetOk("network"); ok {
+			for _, item := range nwSet.(*schema.Set).List() {
+				nwData := item.(map[string]interface{})
+				configNetworkID := util.CastInt64(nwData["network_id"])
```
jooola: Could we be a bit more defensive when casting the now optional `network_id` attribute? See other comment about similar case.

internal/server/resource_test.go
```diff
@@ -699,6 +699,319 @@ func TestAccServerResource_DirectAttachToNetwork(t *testing.T) {
...
+				),
+			},
+			{
+				// Verify no drift when using network_id
+				Config: tmplMan.Render(t,
+					"testdata/r/hcloud_ssh_key", sk,
+					"testdata/r/hcloud_network", nwRes,
+					"testdata/r/hcloud_network_subnet", snwRes,
+					"testdata/r/hcloud_server", sRes,
+				),
```
jooola: Was the goal here to test the migration from using `network_id` to using `subnet_id`?

If yes, I am not sure this test is covering this use case, and I think we should definitely test the migration.

Otherwise, it is unclear to me the purpose of this test. Could you help me out a bit?

internal/server/resource_test.go
```diff
@@ -699,6 +699,319 @@ func TestAccServerResource_DirectAttachToNetwork(t *testing.T) {
...
+			{
+				// Verify no changes detected on subsequent plan
+				Config: tmplMan.Render(t,
+					"testdata/r/hcloud_ssh_key", sk,
+					"testdata/r/hcloud_network", nwRes,
+					"testdata/r/hcloud_network_subnet", snw1Res,
+					"testdata/r/hcloud_network_subnet", snw2Res,
+					"testdata/r/hcloud_server", sRes, // Same config with only subnet_id
+				),
+				PlanOnly: true,
```
jooola: I recall that the terraform acceptance test framework is checking for this already at each step. So this might be unneeded. But please double check, I might be wrong.

jooola (commented):
Here another pass of review. Overrall looks good, but some things could be improved.

Sorry for taking so long to review your PR. The server resource is the most complex resource we have, and we are a bit worried of potentially breaking other users infrastructure. So do not hesitate to add more tests.

internal/server/resource.go
```diff
@@ -204,8 +204,15 @@ func Resource() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"network_id": {
-							Type:     schema.TypeInt,
-							Required: true,
+							Type:       schema.TypeInt,
+							Optional:   true,
+							Computed:   true,
+							Deprecated: "Use subnet_id instead. network_id relies on implicit subnet selection which can lead to unpredictable behavior.",
```
jooola: While we do recommend using the subnet_id attribute, I am unsure about deprecating the network_id attribute. At least if we deprecate it, we should also deprecate it in the `server_network` resource.

But that should be for another PR.

mlinares1998: Agree, i'll remove the deprecated flag then. 😄

internal/server/resource.go
mlinares1998: Working on it.

internal/server/resource.go
mlinares1998: Working on it.

internal/server/resource_test.go
mlinares1998: Seems to be the case, i'll remove this part.

internal/server/resource_test.go
mlinares1998: Indeed, will create a separate test for migration.

mlinares1998 (commented):
@jooola I have improved the PR with your suggestions and added more E2E tests.
Could you give me your feedback 😄 ?

jooola (approved):

jooola:
Thanks for your contribution and your patience!

I'll not release this just yet, and let the e2e tests run a bunch of times on the main branch. I'll try to cut a release #1400 next week if not sooner.

PR #1284, opened 2025-11-27
fix(storage-box): retry snapshot+subaccount create when locked

apricote:
The API only allows users to create a single snapshot or subaccount for any storage box at the same time. While it may be uncommon to create multiple snapshots at the same time, I can see creating many subaccounts at the same time being a common thing.

This uses the default exponential backoff and limit of 5 retries each, so if many resources are "queued", the user may still run into issues.

### DIFF
--- a/internal/storageboxsnapshot/resource.go
+++ b/internal/storageboxsnapshot/resource.go
@@ -134,7 +134,7 @@ func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp 
 	// Create in API
 	// For a single storage box, only a single snapshot can be created simultaneously, all others fail with `locked` error.
 	var result hcloud.StorageBoxSnapshotCreateResult
-	err := control.Retry(control.DefaultRetries, func() error {
+	err := control.Retry(2*control.DefaultRetries, func() error {
 		var err error
 
 		result, _, err = r.client.StorageBox.CreateSnapshot(ctx, storageBox, opts)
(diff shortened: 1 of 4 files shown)


internal/storageboxsnapshot/resource.go
```diff
@@ -131,7 +132,23 @@ func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp
...
+		result, _, err = r.client.StorageBox.CreateSnapshot(ctx, storageBox, opts)
+		if err != nil {
+			if hcloud.IsError(err,
+				hcloud.ErrorCodeLocked,
+			) {
+				return err
+			}
+
+			return control.AbortRetry(err)
+		}
```
jooola, line 149: Would it be possible to wait for existing actions on the Storage Box to finish before continuing?

I think we are missing the call to get actions for a specific storage box, but maybe we could:
- Fetch all actions for a specific storage box
- Filter actions that focus on snapshots/subaccounts
- Settle actions
- Retry

apricote: In the end you would need to do this inside the `retry` call, as other terraform resources could always start new action before you have the chance to start your own.

But yes, this would make the handling a bit cleaner and more reliable, and we would need fewer retries (though each iteration may take longer).

jooola: So nice to have, but for the future.

I think this would be a valid argument for implementing `GET /api/<resource>/<id>/actions` in hcloud-go.

jooola (approved):

PR #1294, opened 2025-12-10
fix(storage-box): run actions serially

apricote:
While the API currently allows triggering multiple actions in parallel for the same Storage Box or Subaccount, this occasionally results in errors. We are opting for reliability in favour of best-case time.


jooola (approved):

apricote:
Failed e2e test is a known flake.

Issue #1287, opened 2025-12-01
[Bug]: StorageBox Error on SSH Keys


tebriel:
### What happened?

Related to #1285 

I have an existing Storage Box, in case of catastrophe when I imported my storage box I also set an SSH key (to help me bootstrap faster) to the imported storage box, however this caused an exception with the provider. (The warning is expected, the error is not)

```
│ Warning: Updating SSH Keys is not possible
│
│   with hcloud_storage_box.data,
│   on storageboxes.tf line 19, in resource "hcloud_storage_box" "data":
│   19:   ssh_keys = [
│   20:     "ssh-ed25519 AAAA_MY_KEY personal"
│   21:   ]
│
│ It is not possible to update the SSH Keys through the API. To avoid accidental data deletion changing the SSH Key in Terraform does not forcibly
│ re-create the Storage Box, but only shows this warning.
│
│ Please use SSH to update the SSH Keys, or manually taint the Storage Box resource so Terraform deletes it and then creates a new one.
╵
╷
│ Error: Provider produced invalid plan
│
│ Provider "provider[\"registry.opentofu.org/hetznercloud/hcloud\"]" planned an invalid value for hcloud_storage_box.data.ssh_keys: planned value
│ cty.NullVal(cty.Set(cty.String)) does not match config value cty.SetVal([]cty.Value{cty.StringVal("ssh-ed25519
│ AAAA_MY_KEY personal")}).
```

### What did you expect to happen?

I expected it to ignore the attribute (like it would if I had set `lifecycle { ignore_changes = [ ssh_keys ] }`

### Please provide a minimal working example

```terraform
import {
  to = hcloud_storage_box.data
  id = "SOME_EXISTING_ID"
}
resource "hcloud_storage_box" "data" {
  name             = "Data"
  storage_box_type = "bx11"
  location         = "fsn1"
  password         = "fakepassword"

  # Ignored after create
  ssh_keys = [
    "ssh-ed25519 AAAA_MY_KEY personal"
  ]

  lifecycle {
    ignore_changes = [
      password
    ]
  }
}
```

tebriel:
note I also tried using `toset` on the ssh keys with `toset(["the key"])` but this failed similarly 

-- edit

[...]

Issue #1285, opened 2025-11-27
Experimental Storage Box Tracker


apricote:
This issue tracks the experimental support for Storage Boxes. Our plan is to remove the experimental marker after a month, but we may extend that if there are open issues that we want to resolve first.

During the experimental support phase we reserve the right to make breaking changes related to the Storage Boxes in minor releases.

apricote:
We implemented a breaking change in the `hcloud_storage_box` resource considering the `ssh_keys` attribute. Please read the upcoming release notes for `v1.58.0` and the related pull request https://github.com/hetznercloud/terraform-provider-hcloud/pull/1296 for details.

You can find a preview of the release notes in #1295.

PR #1296, opened 2025-12-10
feat(storage-box): stop ignoring changes to ssh keys and replace resource instead

apricote:
The current behaviour is not how most other terraform resources work. In general Terraform providers trust that the user reads the proposed plan and understands what they mean. We also have no special protection for other resources that force them to be re-created, like "user_data" or "ssh_keys" on servers.

To aid users I added an example that shows how to set the attribute and how to "safeguard" the resource against accidental deletion.

Another reason to remove this behaviour, is that it is not possible to import resources right now, as Terraform does accept our empty list in face of the users configured SSH Keys as an acceptable state. See #1287 for this error.

---

This is a **breaking change** for the `hcloud_storage_box`. This is accepted in a minor release as the resource is marked as experimental (#1285). I will add a warning to the release notes about this.


tebriel (approved):
based on the limitations for import and changes I think this looks reasonable as a way to communicate to users about the dangers and also give them the freedom/permission to handle resources as is standard. Thanks!

docs/resources/storage_box.md
```diff
@@ -48,6 +48,32 @@ resource "hcloud_storage_box" "backups" {
...
+  ssh_keys = [
+    hcloud_ssh_key.my_key.public_key,
+    file("~/.ssh/id_ed25519.pub"),
+  ]
+
+  lifecycle {
+    # To avoid accidentally deleting the resource you can ignore changes in the ssh_keys attribute:
+    ignore_changes = [
+      ssh_keys
+    ]
```
jooola, line 71: Should we add the `ignore_changes` snippet in all examples? Seems like the default use-case.

jooola (approved):

docs/resources/storage_box.md
apricote: You mean the `backups` example above? I dont think we have any other examples for storage boxes.

For the `backups` example I am not sure if its really relevant, as that examples does not define an SSH Key.

apricote:
Documented the breaking change in the release notes and the experimental tracker issue.

PR #1301, opened 2025-12-13
docs: fix hcloud_zone example

SuperSandro2000:


apricote (changes requested):
Thank you for the fix @SuperSandro2000! Could you `make generate` to also update the "rendered" version of the documentation?

jooola (approved):

PR #1304, opened 2025-12-19
test: remove unused test sweepers

apricote:
The terraform test sweeper framework allows us to define functions that can "clean up" a project after tests, making sure that all temporary resources are deleted. This is only necessary if the actual tests failed to delete the resources.

While we previously had sweepers implementated for most resources, they were never actually used/called. This would require us to add a test main to every package:

```go
import (
  "testing"

  "github.com/hashicorp/terraform-plugin-testing/helper/resource"
)

func TestMain(m *testing.M) {
  resource.TestMain(m)
}
```

And only then are the sweeper flags defined. As we never added this to any of our packages, the sweepers are unused.

The alternative to removing these would be to add the above code, but as no one has used this so far I have opted to remove the sweepers instead.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -14,6 +14,7 @@ linters:
     - errchkjson
     - errorlint
     - exhaustive
+    - forbidigo
     - gocheckcompilerdirectives
     - gochecksumtype
     - gocritic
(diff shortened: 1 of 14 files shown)


jooola (approved):
🧹

PR #1309, opened 2025-12-23
feat: deprecate datacenter in primary ips and servers

apricote:
The `datacenter` attributes is deprecated in `Primary IPs` and `Servers` API resources and will be removed after 1 July 2026. See https://docs.hetzner.cloud/changelog#2025-12-16-phasing-out-datacenters for more details.

Therefor `datacenter` attributes is deprecated in favor of the `location` attribute in the following Terraform resources/datasources:

- `hcloud_server`
- `data.hcloud_server`
- `data.hcloud_servers`
- `hcloud_primary_ip`
- `data.hcloud_primary_ip`
- `data.hcloud_primary_ips`

The `location` attribute already exists for Servers, and was added for Primary IPs.

The `datacenter` attribute will not be updated anymore after it is no longer returned by the API. For existing resources this will keep the previous value, for new resources this will result in an empty string.

Please make sure to upgrade to `v1.58.0+` before the removal date to avoid potential crashes in the provider.

---

Also includes the bump for `hcloud-go/v2` to `v2.33.0`, as that markes the `Datacenter` fields as deprecated.

Replaces #1305

### DIFF
--- a/internal/primaryip/resource.go
+++ b/internal/primaryip/resource.go
@@ -2,10 +2,10 @@ package primaryip
 
 import (
 	"context"
-	"errors"
 	"log"
 	"math/rand"
 	"strconv"
+	"strings"
 
 	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
@@ -22,6 +22,8 @@ const ResourceType = "hcloud_primary_ip"
 
 // Resource creates a new Terraform schema for the hcloud_primary_ip resource.
 func Resource() *schema.Resource {
+	locationAttributes := []string{"location", "datacenter", "assignee_id"}
+
 	return &schema.Resource{
 		CreateContext: resourcePrimaryIPCreate,
 		ReadContext:   resourcePrimaryIPRead,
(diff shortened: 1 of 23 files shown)


internal/primaryip/resource.go
```diff
@@ -41,16 +43,25 @@ func Resource() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			},
+			"location": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ExactlyOneOf: locationAttributes,
+			},
```
jooola, line 52: Datacenter was using `ForceNew`, should this also be set here? Will it block the migration from datacenter to location? Should we write a smart config validation that `RequiresReplace` if the location in the dc name does not match the location name? 

I am also not certain what happens when the datacenter state will be set to an empty string, but the config still refers to a datacenter name value. Will it replace the resource?

internal/primaryip/resource.go
```diff
@@ -98,19 +109,24 @@ func resourcePrimaryIPCreate(ctx context.Context, d *schema.ResourceData, m inte
...
-	default:
+	} else if location, ok := d.GetOk("location"); ok {
+		opts.Location = location.(string)
+	} else if datacenter, ok := d.GetOk("datacenter"); ok {
+		// Backward compatible datacenter argument.
+		// datacenter hel1-dc2 => location hel1
+		parts := strings.SplitN(datacenter.(string), "-", 2)
+
+		if len(parts) != 2 {
+			return diag.Errorf("Datacenter name is not valid, expected format $LOCATION-$DC, but got: %s", datacenter.(string))
```
jooola: Is the `$DC` name explicit/understandable enough?

internal/server/resource.go
```diff
@@ -339,21 +340,28 @@ func resourceServerCreate(ctx context.Context, d *schema.ResourceData, m interfa
...
+			return
+		}
+
+		locationName := parts[0]
+		opts.Location = &hcloud.Location{Name: locationName}
+	}
+	locationName := ""
+	if opts.Location != nil {
+		locationName = opts.Location.Name
 	}
```
jooola, line 361: What happens when the config has `hel1-dc2` but the resulting state for the datacenter is empty?

Also the `locationName` variable name is used twice in 2 different scopes.

jooola (approved):

apricote:
CI failure is unrelated, tests related to this PR all passed.

PR #1316, opened 2026-01-07
feat: wait for storage box running actions before retrying

jooola:
Wait for running actions for the storage box to complete before trying to trigger a new action.

### DIFF
--- a/internal/storageboxsnapshot/resource.go
+++ b/internal/storageboxsnapshot/resource.go
@@ -142,6 +142,14 @@ func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp 
 			if hcloud.IsError(err,
 				hcloud.ErrorCodeLocked,
 			) {
+				// Wait for running actions to complete before trying again
+				actions, actionsErr := r.client.StorageBox.Action.AllFor(ctx,
+					storageBox,
+					hcloud.ActionListOpts{Status: []hcloud.ActionStatus{hcloud.ActionStatusRunning}},
+				)
+				if actionsErr == nil {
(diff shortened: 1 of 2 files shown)


jooola:
Trying out the usability of https://github.com/hetznercloud/hcloud-go/pull/791

internal/storageboxsnapshot/resource.go
```diff
@@ -142,6 +142,14 @@ func (r *Resource) Create(ctx context.Context, req resource.CreateRequest, resp
...
 				hcloud.ErrorCodeLocked,
 			) {
+				// Wait for running actions to complete before trying again
+				actions, actionsErr := r.client.StorageBox.Action.AllFor(ctx,
+					storageBox,
+					hcloud.ActionListOpts{Status: []hcloud.ActionStatus{hcloud.ActionStatusRunning}},
+				)
+				if actionsErr == nil {
+					resp.Diagnostics.Append(hcloudutil.SettleActions(ctx, &r.client.Action, actions...)...)
+				}
```
apricote, line 152: I think we should skip failures of unrelated actions for this resource. These failures should be reported in their own terraform resources. Just retry our own action when the external actions finished/failed.

jooola: Hmm, yes indeed. Maybe this whole feature need some more thought put into it.

I also think that having both the retry loop and the settle actions to wait is a bit redundant. But we probably still want to add some jitter after the actions wait, to prevent 2 racing resources from "unpausing" at the same time.

jooola:
Closing this PR until we find a good way forward.

Issue #1297, opened 2025-12-10
[Bug]: Cannot create hcloud_storage_box_subaccounts


jdelic:
### What happened?

Terraform will try to create storage box subaccounts in parallel. Creating the first one fails for unknown reasons (I can create it on the web interface), but every other one fails because the resource is locked.

```
[...]
Error: API request failed
│ 
│   with hcloud_storage_box_subaccount.serveraccounts["lb1.*****"],
│   on main.tf line 262, in resource "hcloud_storage_box_subaccount" "serveraccounts":
│  262: resource "hcloud_storage_box_subaccount" "serveraccounts" {
│ 
│ An unexpected error was encountered during an API request.
│ 
│ Action is locked
│ 
│ Error code: locked
│ Status code: 423
│ 
╵
╷
│ Error: Action failed
│ 
│   with hcloud_storage_box_subaccount.serveraccounts["apps1.*****"],
│   on main.tf line 262, in resource "hcloud_storage_box_subaccount" "serveraccounts":
│  262: resource "hcloud_storage_box_subaccount" "serveraccounts" {
│ 
│ An API action for the resource failed.
│ 
│ creating subaccount failed
│ 
│ Error code: action_failed
│ Command: create_subaccount
│ ID: 598022669323684
│ Resources: storage_box: 500400, storage_box_subaccount: 159164
```

### What did you expect to happen?

I expected around 5 subaccounts to be created.

### Please provide a minimal working example

Here is the code snippet that creates this error:

```hcl
locals {
    server_config = {
        "db.local" = {
            server_type = "cx22"
        }
        "test.local" = {
            server_type = "cx22"
        }
        "test2.local" = {
            server_type = "cx32"
        }
    }
}
        
resource "random_password" "storage_box_root" {
    length = 32
    special = true
    override_special = "$%+-#"
    min_special = 1
    upper = true
    min_upper = 1
    lower = true
    min_lower = 1
    min_numeric = 1
}

resource "hcloud_storage_box" "backup-box" {
    name = "backup-box"
    storage_box_type = "bx11"
    location = "hel1"

    password = random_password.storage_box_root.result

    access_settings = {
        ssh_enabled = true
    }

    ssh_keys = [hcloud_ssh_key.ssh1]
    delete_protection = false

    depends_on = [hcloud_ssh_key.ssh1]
}


resource "random_password" "backup_accounts" {
    for_each = local.server_config

    length = 32
    special = true
[...]

PR #1331, opened 2026-01-23
docs: Storage Box Subaccount home_directory should not starting with a /

jdelic:
Currently that will fail in the API with a generic error message. (see #1297 for details)


jooola:
Thanks for the contribution!

jooola (approved):

PR #1195, opened 2025-09-25
fix: wait for floating_ip assign action to complete

jooola:
Ensures that the Floating IP assign background task succeeds by waiting on the returned action. Report an error if the action failed.

PR #1335, opened 2026-01-28
feat(floating-ip): unassign from server before deleting it

apricote:
Similar to the behaviour of Primary IPs, make sure that the Floating IP is correctly unassigned before deleting it.

### DIFF
--- a/internal/floatingip/resource.go
+++ b/internal/floatingip/resource.go
@@ -250,9 +250,36 @@ func resourceFloatingIPDelete(ctx context.Context, d *schema.ResourceData, m int
 		d.SetId("")
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


jooola (approved):

jooola:
Hmm, CI is failing:
```
    resource_assignment_test.go:64: Error running post-test destroy, there may be dangling resources: exit status 1
        
        Error: Floating IP with ID 115181381 is not assigned to any resource (service_error, a42af560904b7bd703863609c33caee8)
        
--- FAIL: TestAccFloatingIPAssignmentResource (80.98s)
```

jooola:
I had a similar problem in #1195, maybe linter is a good idea.

PR #1380, opened 2026-04-10
fix(firewall): apply_to update removes firewall from state when target resource is not found

### LINKED ISSUE #1379: [Bug]: hcloud_firewall apply_to update incorrectly removes firewall from state when target resource is not found
### What happened?

When replacing a server that has a firewall applied via `apply_to`, the firewall resource is incorrectly removed from state. This produces the error:

```
Error: Provider produced inconsistent result after apply

When applying changes to hcloud_firewall.this, provider
"provider[\"registry.opentofu.org/hashicorp/hcloud\"]" produced an
unexpected new value: root object was present, but now absent.

This is a bug in the provider, which should be reported in the provider's
own issue tracker.
```

The root cause is in `syncApplyTo` (`internal/firewall/resource.go`). When `Firewall.RemoveResources` is called to detach the old (already destroyed) server, the Hetzner API returns a generic `not_found` error. The error handler calls `resourceFirewallIsNotFound`, which calls `d.SetId("")` — removing the firewall from state, even though it's the server that does not exist anymore.

This also means that the firewall is never applied to the replacement server.

The firewall should remain in state and be correctly applied to the replacement target.

```hcl
resource "hcloud_server" "this" {
  name        = "example"
  image       = "ubuntu-24.04"
  server_type = "cx23"
}

resource "hcloud_firewall" "this" {
  name = "example"

  apply_to {
    server = hcloud_server.this.id
  }
}
```

```shell
tofu apply
tofu apply -replace="hcloud_server.this"
```

The second apply produces the error above.

jochemloedeman:
In `syncApplyTo`, replace the `resourceFirewallIsNotFound` call with a direct `hcloud.IsError(err, hcloud.ErrorCodeNotFound)` check that logs a warning and continues execution instead of removing the firewall from state and returning early.

Closes #1379

### DIFF
--- a/internal/firewall/resource.go
+++ b/internal/firewall/resource.go
@@ -379,12 +379,11 @@ func syncApplyTo(ctx context.Context, d *schema.ResourceData, client *hcloud.Cli
 	if len(removeResources) > 0 {
 		actions, _, err := client.Firewall.RemoveResources(ctx, firewall, removeResources)
 		if err != nil {
-			if resourceFirewallIsNotFound(err, d) {
-				return nil
+			if !hcloud.IsError(err, hcloud.ErrorCodeNotFound) {
+				return hcloudutil.ErrorToDiag(err)
 			}
-			return hcloudutil.ErrorToDiag(err)
-		}
-		if err := waitForFirewallActions(ctx, client, actions, firewall); err != nil {
+			log.Printf("[WARN] firewall (%s) resource not found, skipping remove: %v", d.Id(), err)
+		} else if err := waitForFirewallActions(ctx, client, actions, firewall); err != nil {
 			return hcloudutil.ErrorToDiag(err)
 		}
 	}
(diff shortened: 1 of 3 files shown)


jooola:
Hey, thanks for the contribution, could you add some acceptance test to make sure we don't break this again?

jooola:
The e2e test failures in the CI are expected.

jochemloedeman:
> Hey, thanks for the contribution, could you add some acceptance test to make sure we don't break this again?

Done. On main, `TestAccFirewallResource_ApplyToServerReplace` gives
```
resource_test.go:207: Step 2/2 error: Error running apply: exit status 1
    
    Error: Provider produced inconsistent result after apply
    
    When applying changes to hcloud_firewall.applyto-firewall, provider
    "provider[\"registry.terraform.io/hashicorp/hcloud\"]" produced an unexpected
    new value: Root object was present, but now absent.
    
    This is a bug in the provider, which should be reported in the provider's own
    issue tracker.
```
It passes on this branch :)

jochemloedeman:
@jooola let me know if you need anything else!

jooola:
I took another look at this PR and it all looks great!

But I found a small bug in the API, and I'd like to tackle it before merging this.

Thanks for your patience :smile:

jooola:
Ran the e2e tests locally, everything is green. Thanks for the contribution!

jooola (approved):

PR #1395, opened 2026-04-28
feat(primary-ip): `assignee_type` attribute is now optional

lukasmetzner:
- Raise warning if `assignee_type` is configured without `assignee_id`
- Raise error if `assignee_id` is configured without `assignee_type`

#### Primary IPs `assignee_type` behavior change

As of 1 August 2026, the behavior of the Primary IP `assignee_type` property will change, and will return `unassigned` when the Primary IP is not assigned (when `assignee_id` is `null`). The goal is to eventually assign Primary IPs to other resource types, not only to `server`.

See the [changelog](https://docs.hetzner.cloud/changelog#2026-04-27-primary-ips-will-return-unassigned) for more details.

In addition, the Primary IP request body `assignee_type` property of the operation [`POST /v1/primary_ips`](https://docs.hetzner.cloud/reference/cloud#tag/primary-ips/create_primary_ip) is now optional. Primary IPs created without `assignee_type` return `server` until 1 August 2026, after this date, its value will be `unassigned`.

See the [changelog](https://docs.hetzner.cloud/changelog#2026-04-27-primary-ips-make-assignee_type-optional) for more details.

### DIFF
--- a/docs/resources/primary_ip.md
+++ b/docs/resources/primary_ip.md
@@ -38,28 +38,26 @@ communicating with the API.
 
 ```terraform
 resource "hcloud_primary_ip" "main" {
-  name          = "primary_ip_test"
-  location      = "fsn1"
-  type          = "ipv4"
-  assignee_type = "server"
-  auto_delete   = true
+  name        = "primary-ip"
+  location    = "fsn1"
+  type        = "ipv4"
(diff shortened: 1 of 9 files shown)


jooola (approved):
E2e test fail on unrelated resources.

Issue #1005, opened 2024-10-02
Support for s3 object storage


KiaraGrouwstra:
### What whould you like to see?

hetzner recently introduced their S3-compatible [object storage](https://docs.hetzner.com/storage/object-storage/overview/), offering immutable storage cheaper than their regular shared volumes.
it would be cool if this provider would facilitate configuring hetzner object storage as well, tho it seems given it's in beta there is currently still a manual step to request access involved as well.

apricote:
Hey @KiaraGrouwstra,

right now we do not plan to add support for the Object Storage S3 API in this terraform provider. You can use any S3-compatible provider like the [Minio provider](https://docs.hetzner.com/storage/object-storage/getting-started/creating-a-bucket-minio-terraform) instead.

If using other providers does not work for you, could you explain the issues you have with them and the benefits you see with adding the APIs to this provider?

KiaraGrouwstra:
i'll try that one - thank you for your response!

c33s:
i would have prefered to have all from one hand. it would also feel strange if i could create server, firewalls, ... via hetzner cli but not the object storage.
also s3 user and secret management would be nice to have in this provider.

i would vote for reopening the issue

BerndDA:
As I understood, creating buckets and access keys must be done using Hetzner API. We already successfully configured existing buckets (object lifecycle ruels) using other existing providers, but it would be nice to be able to create the buckets using this provider (saves a manual step in the UI)

jooola:
Hello all 👋

All our integrations rely on the Hetzner Cloud public API, which is available with a certain level of stability. Since the features you are requesting are not in the public API, we cannot implement them.

Therefore, **for the time being**, we do not plan to support:

- Generating credentials using the Public API.
- Creating buckets using the Public API: You can use [an S3-compatible tool](https://docs.hetzner.com/storage/object-storage/getting-started/using-s3-api-tools) instead, for example using [curl](https://docs.hetzner.com/storage/object-storage/getting-started/using-curl) or the [minio terraform provider](https://docs.hetzner.com/storage/object-storage/getting-started/creating-a-bucket-minio-terraform).
- Embedding a S3 client in our integrations: You can use [any S3-compatible tool](https://docs.hetzner.com/storage/object-storage/getting-started/using-s3-api-tools) instead.

Note that only a [subset of the Amazon S3 features are currently supported](https://docs.hetzner.com/storage/object-storage/supported-actions).

[...]

PR #1402, opened 2026-05-07
docs(s3): add reference to S3 guide

nachtjasmin:
By adding a reference to the existing guide on docs.hetzner.com, customers that want to manage the S3 buckets declaratively no longer have to search for #1005 or for said documentation via their preferred search engine.

## Open dicussion points

1. Should we add a reference to the GitHub issue?
2. Should we write down more details about why the feature isn't there, effectively duplicating some of the information in the GitHub issue for broader transparency?

### DIFF
--- a/templates/guides/s3-object-storage.md
+++ b/templates/guides/s3-object-storage.md
@@ -1,0 +1,11 @@
+---
+subcategory: ""
+page_title: "Use Hetzner Object Storage (S3)"
+---
+
+As of today, there's no native Hetzner Cloud API to manage the S3 buckets and credentials.
+Therefore, the only supported method for managing the resources of a Hetzner Object Storage is via third-party providers.
+
+We provide an [example workflow for creating new buckets using the `aminueza/minio` provider in the Hetzner documentation](https://docs.hetzner.com/storage/object-storage/getting-started/creating-a-bucket-minio-terraform/).
+
+More information about the missing support can be found in [GitHub issue #1005](https://github.com/hetznercloud/terraform-provider-hcloud/issues/1005).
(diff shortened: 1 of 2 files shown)


templates/guides/s3-object-storage.md
```diff
@@ -0,0 +1,23 @@
+---
+subcategory: ""
+page_title: "Use Hetzner Object Storage (S3)"
+---
+
+# Use Hetzner Object Storage (S3) with the `aminueza/minio` provider
```
jooola: I think we can add some docs about the object storage support, but I would prefer to keep it generic (e.g. not mention any providers). We do not want to monitor the upstream docs to make sure everything is up to date.

templates/guides/s3-object-storage.md
```diff
@@ -0,0 +1,23 @@
...
+
+To import existing S3 buckets to your state, you must use the name of the bucket as the `id`.
+Example `import` block:
+
+```hcl
+import {
+  to = minio_s3_bucket.my_bucket
+  id = "my-bucket-h7892dd9"
+}
+```
```
jooola: This probably belongs to the upstream docs.

jooola (commented):
Good idea, feel free to also link Github issue so people can dig a bit more if needed.

With this in mind, we don't necessarily need to give more details, if needed people can browse to the linked issue.

templates/guides/s3-object-storage.md
nachtjasmin: Updated it, now it only links to the upstream docs.

templates/guides/s3-object-storage.md
nachtjasmin: Will prepare a separate MR internally. :+1:

PR #1342, opened 2026-02-09
feat(server): implement actions for poweron, poweroff, reset, and reboot

DanielMSchmidt:
Terraform introduced [actions](https://danielmschmidt.de/posts/2025-09-26-terraform-actions-introduction/) in v 1.14 and I was thinking an action to power on, power off, reboot, and reset a machine would make for a nice addition. I personally want to host a development setup on Hetzner and being able to power down machines that I don't need would be quite convenient.

PR #1410, opened 2026-05-20
feat: Add declarative power management for hcloud_server

wizenink:
Adds a new `hcloud_server_power_state` SDK resource to manage Hetzner Cloud server power state declaratively through normal plan/apply workflows.

This enables OpenTofu-compatible power on/off management without relying on Terraform actions, while keeping the existing action-based resources unchanged. The new resource supports `state = "running"` and `state ="off"`,  reports the raw server status, and removes only power-state management on destroy without changing the server.

### DIFF
--- a/docs/resources/server_power_state.md
+++ b/docs/resources/server_power_state.md
@@ -1,0 +1,49 @@
+---
+page_title: "hcloud_server_power_state Resource - hcloud"
+subcategory: ""
+description: |-
+  Manages the power state of a Hetzner Cloud Server.
+---
+
+# hcloud_server_power_state
+
+Manages the power state of a Hetzner Cloud Server.
+
+This resource only manages whether the server is running or off. Destroying this resource removes power state management from Terraform/OpenTofu state and does not power off, power on, or delete the server.
+
+## Example Usage
+
+```terraform
+resource "hcloud_server" "example" {
+  name        = "example"
+  server_type = "cx22"
+  image       = "debian-12"
+}
+
+resource "hcloud_server_power_state" "example" {
+  server_id = hcloud_server.example.id
+  state     = "off"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `server_id` - (Required, int) ID of the server whose power state should be managed.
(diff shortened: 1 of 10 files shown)


jooola:
Does https://github.com/hetznercloud/terraform-provider-hcloud/pull/1342 solve the problem you are trying to solve with this PR?

wizenink:
@jooola not exactly. Actions are only supported on Hashicorp's Terraform, not on OpenTofu, so tofu does currently not have a way to change the power state declaratively. This PR aims to backport some of that into the non-action way of doing so

jooola:
Hmm, I am not sure this is a good idea. If we are trying to go faster than opentofu, maybe the work effort should be directed towards supporting `actions` in opentofu directly: https://github.com/opentofu/opentofu/issues/3309

I'll discuss this addition internally and come back to you.

wizenink:
I don't find it conflicting. IMO actions should be an adition into opentofu, but declarative old-fashioned way is still nice for some use cases. f.ex AWS and Azure can use both ways. Anyway, let me know and I'll maybe work on the opentofu path if I'm available. Thank you!

jooola:
@wizenink Could you please explain and describe your use-case in https://github.com/hetznercloud/terraform-provider-hcloud/issues/1414 so we better understand the reason behind this feature request?

nh-exa:
I am also very interested in this feature. It would allow us to start and stop servers via simple state changes. AWS and Azure already have this.

nh-exa:
I'd still like to see this feature. Any progress here @jooola @wizenink ?

jooola:
@nh-exa Please describe your use case in details in #1414  or add a :+1: 

With regards to this PR, we will not be able to merge this as it is. A lot of work is still needed to support managing the state of the server:
- Migrate the server resource to the plugin framework (#752)
- Ensure the server resource does not start/stop a server when not desired, the resource currently always try to start the server before finishing.
- Find the best design based on the feedback collected in #1414 

So I am happy to leave this PR open until then, but we will not spend time reviewing this before the above points have not been solved.

Issue #1136, opened 2025-06-05
[Bug]: data.hcloud_load_balancer has null `service` even though there are services


Preisschild:
### What happened?

When I use the `hcloud_load_balancer` datasource the `service` list is set to `null` even though there are two services (shown on web interface & hcloud cli).

The other properties (like `target`) are set properly.

### What did you expect to happen?

To see the two Load Balancer services in `data.hcloud_load_balancer.<name>.service`

### Please provide a minimal working example

```terraform
data "hcloud_load_balancer" "lb" {
  name = "myloadbalancer"
}

output "hcloud_load_balancer" {
  value = data.hcloud_load_balancer.lb
}
```

PR #1426, opened 2026-05-28
feat(load-balancer): add service data source

### LINKED ISSUE #1413: [Bug]: `hcloud_load_balancer` data source not correctly populated
The `hcloud_load_balancer` data source does not populate the following attributes correctly:
- `service` is missing entirely
- `network_id` and `network_ip` only ever show the values of the first attached private network

phm07:
This PR adds a dedicated data source for Load Balancer Services. The `services` field in the original Load Balancer data source was not populated and is deprecated after this PR.

Fixes #1413 and #1136


internal/loadbalancer/data_source_service.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/hetznercloud/terraform-provider-hcloud/internal/util/hcloudutil"
+)
+
+// DataSourceServiceType is the type name of the Hetzner Cloud Load Balancer Service resource.
+const DataSourceServiceType = "hcloud_load_balancer_service"
+
+func getCommonServiceDataSchema(readOnly bool) map[string]schema.Attribute {
+	return map[string]schema.Attribute{
+		"id": schema.StringAttribute{
+			Optional: !readOnly,
```
jooola: Please write the description for all attributes (`MarkdownDescription`).

internal/loadbalancer/data_source_service.go
```diff
@@ -0,0 +1,240 @@
...
+				"cookie_name": schema.StringAttribute{
+					Computed: true,
+				},
+				"cookie_lifetime": schema.Int32Attribute{
+					Computed: true,
+				},
+				"certificates": schema.ListAttribute{
+					Computed:    true,
+					ElementType: types.Int64Type,
+				},
```
jooola: Because this is a new resource, we might want to rename a few misnamed arguments. This should probably have an _id(s) suffix.

internal/loadbalancer/data_source_service_list.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// ConfigValidators returns a list of ConfigValidators. Each ConfigValidator's Validate method will be called when validating the data source.
+func (d *DataSourceServiceList) ConfigValidators(_ context.Context) []datasource.ConfigValidator {
+	return []datasource.ConfigValidator{
+		datasourcevalidator.ExactlyOneOf(
+			path.MatchRoot("load_balancer_id"),
+		),
+	}
+}
```
jooola: Not needed, the argument is required already.

internal/loadbalancer/data_source_service_list.go
```diff
@@ -0,0 +1,147 @@
...
+
+// Schema should return the schema for this data source.
+func (d *DataSourceServiceList) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema.MarkdownDescription = "Provides a list of Hetzner Cloud Load Balancer Services."
+
+	resp.Schema.Attributes = map[string]schema.Attribute{
+		"id": schema.StringAttribute{
+			Computed: true,
+		},
```
jooola: I think this is not needed, those were only kept for backward compatibility in other data sources.

internal/loadbalancer/model.go
```diff
@@ -37,3 +40,117 @@ func populateNetworkResourceData(
...
+	m.Proxyprotocol = types.BoolValue(hc.Proxyprotocol)
+	m.Protocol = types.StringValue(string(hc.Protocol))
+	m.HTTP = types.ObjectValueMust(m.tfAttributesTypesHTTP(), map[string]attr.Value{
+		"sticky_sessions": types.BoolValue(hc.HTTP.StickySessions),
+		"cookie_name":     types.StringValue(hc.HTTP.CookieName),
+		"cookie_lifetime": types.Int32Value(util.CastInt32(hc.HTTP.CookieLifetime.Seconds())),
+		"certificates":    types.ListValueMust(types.Int64Type, httpCertIDs),
+		"redirect_http":   types.BoolValue(hc.HTTP.RedirectHTTP),
+		"timeout_idle":    types.Int32Value(util.CastInt32(hc.HTTP.TimeoutIdle.Seconds())),
+	})
```
jooola: Please use a new struct for nested objects. See https://github.com/hetznercloud/terraform-provider-hcloud/blob/server-test-coverage/internal/storagebox/model.go#L99 as example.

internal/loadbalancer/model.go
```diff
@@ -37,3 +40,117 @@ func populateNetworkResourceData(
...
+var _ util.ModelToTerraform[types.Object] = &serviceModel{}
+
+func (m *serviceModel) FromAPI(_ context.Context, hc *hcloud.LoadBalancerService) diag.Diagnostics {
+	var httpCertIDs []attr.Value
+	for _, cert := range hc.HTTP.Certificates {
+		httpCertIDs = append(httpCertIDs, types.Int64Value(cert.ID))
+	}
+
+	m.ListenPort = types.Int32Value(util.CastInt32(hc.ListenPort))
+	m.DestinationPort = types.Int32Value(util.CastInt32(hc.DestinationPort))
```
jooola: Do we really need CastInt32?

Because the data is coming from the API, I'd expect the data to be valid and not be bigger than max int32.

jooola (commented):
Did a first pass, looks good. It needs some more work though.

internal/loadbalancer/model.go
phm07: I added it because the linter was complaining about the `int32()` cast, so I added a "proper" method. We can also just use the builtin cast and ignore the linter error.

internal/loadbalancer/data_source_service.go
```diff
@@ -0,0 +1,265 @@
...
+		"id": schema.StringAttribute{
+			MarkdownDescription: "ID of the Load Balancer Service. Format: `<load_balancer_id>__<listen_port>`",
+			Optional:            !readOnly,
+			Computed:            true,
+		},
+		"load_balancer_id": schema.Int64Attribute{
+			MarkdownDescription: "ID of the Load Balancer this Service belongs to.",
+			Optional:            !readOnly,
+			Computed:            true,
+		},
```
jooola, line 29: This is required for the "single" data source.

internal/loadbalancer/data_source_service.go
```diff
@@ -0,0 +1,265 @@
...
+// DataSourceServiceType is the type name of the Hetzner Cloud Load Balancer Service resource.
+const DataSourceServiceType = "hcloud_load_balancer_service"
+
+func getCommonServiceDataSchema(readOnly bool) map[string]schema.Attribute {
+	return map[string]schema.Attribute{
+		"id": schema.StringAttribute{
+			MarkdownDescription: "ID of the Load Balancer Service. Format: `<load_balancer_id>__<listen_port>`",
+			Optional:            !readOnly,
+			Computed:            true,
+		},
```
jooola, line 24: I think this can be marked as only computed. If the users have the service ID, its probably because they already have a resource for that service. In that case, their is no point in using the data source. 

And I find it fragile to ask the users to build the id by hand, we might as well require the user to pass the listen port and lb id directly.

internal/loadbalancer/data_source_service.go
```diff
@@ -0,0 +1,265 @@
...
+
+		listenPort := int(data.ListenPort.ValueInt32())
+		for _, _svc := range lb.Services {
+			if _svc.ListenPort == listenPort {
+				svc = &_svc
+				break
+			}
+		}
+		if svc == nil {
+			resp.Diagnostics.Append(hcloudutil.NotFoundDiagnostic("load_balancer_service", "listen_port", data.ListenPort.String()))
```
jooola:
```suggestion
			resp.Diagnostics.Append(hcloudutil.NotFoundDiagnostic("load balancer service", "listen_port", data.ListenPort.String()))
```

This is usually a human readable string.

internal/loadbalancer/resource_service.go
```diff
@@ -348,15 +348,26 @@ func resourceLoadBalancerServiceDelete(ctx context.Context, d *schema.ResourceDa
 	return nil
 }
 
-func setLoadBalancerServiceSchema(d *schema.ResourceData, lb *hcloud.LoadBalancer, svc *hcloud.LoadBalancerService) {
+func setLoadBalancerServiceSchema(d *schema.ResourceData, lb *hcloud.LoadBalancer, svc *hcloud.LoadBalancerService, idAsString bool) {
+	util.SetSchemaFromAttributes(d, getLoadBalancerServiceAttributes(lb, svc, idAsString))
+}
```
jooola: Are the changes in this file still needed? What are you fixing?

phm07: No, the changes are left over, I undid them

internal/loadbalancer/model.go
```diff
@@ -37,3 +41,208 @@ func populateNetworkResourceData(
...
+	m.Proxyprotocol = types.BoolValue(hc.Proxyprotocol)
+	m.Protocol = types.StringValue(string(hc.Protocol))
+
+	{
+		value := serviceModelHTTP{}
+		diags.Append(value.FromAPI(ctx, &hc.HTTP)...)
+
+		m.HTTP, newDiags = value.ToTerraform(ctx)
+		diags.Append(newDiags...)
+	}
```
jooola: Because the HTTP model is only set when the protocol is `http` or `https`, we must handle the case when it is unset.

We must write something into the state, we cannot have attributes that are unknown.

With this in mind, we should check for the protocol and only populate the server http model when relevant, otherwise explicitly save a null object: `m.HTTP = types.ObjectNull((&serviceModelHTTP{}).tfAttributesTypes())`

internal/loadbalancer/model.go
```diff
@@ -37,3 +41,208 @@ func populateNetworkResourceData(
...
+	m.Timeout = types.Int32Value(int32(hc.Timeout.Seconds()))
+	m.Retries = types.Int32Value(int32(hc.Retries)) //nolint:gosec
+
+	if hc.HTTP != nil {
+		value := serviceModelHealthCheckHTTP{}
+		diags.Append(value.FromAPI(ctx, hc.HTTP)...)
+
+		m.HTTP, newDiags = value.ToTerraform(ctx)
+		diags.Append(newDiags...)
+	}
```
jooola, line 205: Same as above, we must store something, an unknown attribute is not gonna work.

In the else scope, could you add to save a null object?

internal/loadbalancer/data_source_service.go
```diff
@@ -0,0 +1,233 @@
...
+		err error
+	)
+
+	switch {
+	case !data.ID.IsNull():
+		lb, svc, err = lookupLoadBalancerServiceID(ctx, data.ID.ValueString(), d.client)
+		if err != nil {
+			resp.Diagnostics.Append(hcloudutil.APIErrorDiagnostics(err)...)
+			return
+		}
```
jooola: We do not support fetching by ID only. Please remove.

internal/loadbalancer/data_source_service.go
```diff
@@ -0,0 +1,233 @@
...
+
+	case !data.LoadBalancerID.IsNull() && !data.ListenPort.IsNull():
```
jooola: Those are marked as required in the schema, so I think we can assume the values are always set.

internal/loadbalancer/data_source_service.go
```diff
@@ -0,0 +1,233 @@
...
+		for _, _svc := range lb.Services {
+			if _svc.ListenPort == listenPort {
+				svc = &_svc
+				break
+			}
+		}
+		if svc == nil {
+			resp.Diagnostics.Append(hcloudutil.NotFoundDiagnostic("load balancer service", "listen_port", data.ListenPort.String()))
+			return
+		}
```
jooola:
```suggestion
		index := slices.IndexFunc(lb.Services, func(o hcloud.LoadBalancerService) bool {
			return o.ListenPort == int(data.ListenPort.ValueInt32())
		})
		if index < 0 {
			resp.Diagnostics.Append(hcloudutil.NotFoundDiagnostic("load balancer service", "listen_port", data.ListenPort.ValueInt32()))
			return
		}
		svc = &lb.Services[index]
```
Maybe something like this?

internal/loadbalancer/data_source_service.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	if lb == nil || svc == nil {
+		return
+	}
```
jooola: Not needed, since we already handle it above.

internal/loadbalancer/data_source_service_list.go
```diff
@@ -0,0 +1,118 @@
...
+		resp.Diagnostics.Append(hcloudutil.NotFoundDiagnostic("load balancer", "id", data.LoadBalancerID.String()))
+		return
+	}
+
+	var services []*hcloud.LoadBalancerService
+	for _, svc := range lb.Services {
+		services = append(services, &svc)
+	}
+
+	resp.Diagnostics.Append(populateDataSourceServiceListModel(ctx, &data, lb, services)...)
```
jooola: Why are we using a pointer for LoadBalancerService ? Is it really needed?

jooola (commented):
A bunch of things to address and the tests are missing.

For the tests, please try to test fetching the 3 protocols (tcp, http, https).

PR #1427, opened 2026-05-28
fix(server): handle ipv6 when removing public net block

jooola:
The helper function was inlined to simplify and reduce indirection in the code. This also allow us to handle the ipv6 case which wasn't handled before.


lukasmetzner (approved):
Do we have an existing test, where we can add an IPv6 case? Otherwise, looks good 👍

jooola:
Not yet, its being worked on in https://github.com/hetznercloud/terraform-provider-hcloud/pull/1271

lukasmetzner:
> Not yet, its being worked on in #1271

Sound good :rocket:

PR #1429, opened 2026-05-28
feat(primary-ip): make `auto_delete` attribute optional

jooola:
The `auto_delete` attribute is currently required, but we recommend always setting its value to `false`:

     Setting `auto_delete` to `false` is recommended, because if a server assigned to the managed ip
     is getting deleted, it will also delete the primary IP which will break the terraform state.

This change makes the `auto_delete` attribute optional, and defaults to false. Simplifying user's configurations and using a recommended default.

https://docs.hetzner.cloud/reference/cloud#tag/primary-ips/create_primary_ip

### DIFF
--- a/internal/primaryip/resource.go
+++ b/internal/primaryip/resource.go
@@ -126,6 +126,7 @@ communicating with the API.
 			MarkdownDescription: "Whether auto delete is enabled. Setting `auto_delete` to `false` is recommended, because if a server assigned to the managed ip is getting deleted, it will also delete the primary IP which will break the terraform state.",
 			Optional:            true,
 			Computed:            true,
+			Default:             booldefault.StaticBool(false),
 		},
 		"labels": resourceutil.LabelsSchema(),
 		"delete_protection": schema.BoolAttribute{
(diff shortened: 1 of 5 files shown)


internal/primaryip/resource.go
```diff
@@ -124,7 +124,9 @@ communicating with the API.
 		},
 		"auto_delete": schema.BoolAttribute{
 			MarkdownDescription: "Whether auto delete is enabled. Setting `auto_delete` to `false` is recommended, because if a server assigned to the managed ip is getting deleted, it will also delete the primary IP which will break the terraform state.",
```
lukasmetzner: If it defaults to `false`, then `Setting auto_delete to false is recommended, …` feels a bit odd. Should we maybe just state what happens when it is set to `true`? Or recommend to not set the attribute.

lukasmetzner (approved):
Little nit on the docs. Otherwise, looks good to me 🚀

internal/primaryip/resource.go
jooola: Fixed

PR #1443, opened 2026-06-03
feat(server): remove `alias_ips` default value from `hcloud_server_network`

### LINKED ISSUE #1441: [Bug]: Got "Provider produced inconsistent result after apply" for  hcloud_server_network resource
### What happened?

Hello,

I'm using provider version 1.64 and terraform v1.12.0.

My terraform code creates a hcloud_server and a hcloud_server_network resources.
The hcloud_server_network resource looks like this:

```
resource "hcloud_server_network" "server1_in_my_network" {
  network_id = data.hcloud_network.my_network.id
  server_id = hcloud_server.server1.id
  ip = var.server_ip
  lifecycle {
    ignore_changes = [alias_ips]
  }
}
```

The server is a linux deployed from a custom image with user data.
One of the user-data script assigns an alias IP address as part of the cloud-init process.
It seems that cloud-init completes before the creation of the hcloud_server_network resource,
because I observe the following output:
```
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to hcloud_server_network.server1_in_my_network, provider "provider[\"registry.terraform.io/hetznercloud/hcloud\"]"
│ produced an unexpected new value: .alias_ips: actual set element cty.StringVal("<obfuscated ip>") does not correlate with any element in plan.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to hcloud_server_network.server1_in_my_network, provider "provider[\"registry.terraform.io/hetznercloud/hcloud\"]"
│ produced an unexpected new value: .alias_ips: length changed from 0 to 1.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

The hcloud_server_network resource is created 
and the server is attached to the network properly.
I can connect to the server.

I expect the creation of the hcloud_server_network to complete without errors,
because the alias_ips attribute is listed in the ignore_changes section of the lifecycle of the resource

Check the code snippet I provided in the problem description.

jooola:
Some users seem to manage their `alias_ips` externally, so we do not want to assume the list of alias ips is going to be empty.

Fixes #1441

### DIFF
--- a/internal/server/resource_network.go
+++ b/internal/server/resource_network.go
@@ -15,7 +15,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
-	"github.com/hashicorp/terraform-plugin-framework/resource/schema/setdefault"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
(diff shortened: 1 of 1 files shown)


jooola:
In terraform, with the new plugin framework we now have the option to have attributes managed by TF and other managed "externally". We do not yet have a clear separation of which attributes are/should always managed by terraform, and which may be managed externally.

In this PR, we try to change a attribute fully managed by Terraform to be only managed when the user has it in their configuration. This is probably a breaking change, and we will reevaluate this for the next major release.

The issue #1441 also got closed and the author is looking for an alternative way to work around this.

PR #1447, opened 2026-06-04
refactor: allow to get one or get first from list in data source queries

jooola:
When the data source list query return an invalid list of resources, we want to be able to report enough information to the users about the query.

Those helper provide generic helpers to pick one or the first item in the list.


jooola:
Future improvement for those helpers could be to include parent resources in the diagnostics e.g. `hcloudutil.WithParentResource("zone", zone.IDOrName)`.

PR #1448, opened 2026-06-04
refactor(image): move image data sources to Plugin Framework

jooola:
Related to #752


apricote (approved):
🚀

internal/image/data_source.go
```diff
@@ -2,308 +2,296 @@ package image
...
+		"with_selector": schema.StringAttribute{
+			MarkdownDescription: "Filter results using a [Label Selector](https://docs.hetzner.cloud/reference/hetzner#label-selector).",
+			Optional:            true,
+		},
+		"with_status": schema.SetAttribute{
+			MarkdownDescription: "Filter results by statuses, for example `creating` or `available`.",
+			Optional:            true,
+			ElementType:         types.StringType,
+		},
+		"with_architecture": schema.StringAttribute{
```
apricote, line 122: `with_architecture` was a `set` before, changing it to a string is a breaking change (and e2e tests complain).

apricote (changes requested):

internal/image/data_source.go
jooola: Fixed

jooola:
The upgrade tests are failing, because of a difference between the versions. Some of the diffs are expected, but some other I don't know if we want fix the values or stay backward compatible at all costs.

```
// Image list
                  ~ images             = [
                      ~ {
                          ~ deprecated   = "" -> null
                            id           = 40093059
                            name         = "lamp"
                          - selector     = ""
                            # (8 unchanged attributes hidden)
                        },

// Image single
                    id                 = 161547269
                  ~ include_deprecated = false -> null
                    name               = "ubuntu-24.04"
                  ~ with_architecture  = "x86" -> null
                    # (13 unchanged attributes hidden)
```

For the users provided options, I think its fine to change them.

For the `deprecated` attribute I am unsure if we still want to roll with this fix.

apricote:
I think the shown changes are fine, all are basically the same in the terraform type system.

For the `with_architecture` diff: It looks to me like you missed the default in the new schema. This should be set for backwards compatibility.

jooola:
> For the with_architecture diff: It looks to me like you missed the default in the new schema. This should be set for backwards compatibility.

Data sources do not support defaults. I was only able to document the defaults in the markdown description, and set the default in the code directly.

I'd argue because it is a user input used to filter the result, it is fine to have null when it is unset.

apricote (approved):

PR #1475, opened 2026-07-02
chore(ci): support running acceptance tests from forks

nikolauspschuetz:
## What / why

The `Test` workflow's `e2e` job gets its Hetzner Cloud credentials from [`tps-action`](https://github.com/hetznercloud/tps-action), which mints a temporary token from TPS using **GitHub OIDC**. That OIDC identity is only trusted for `hetznercloud/terraform-provider-hcloud`, so **external contributors can't run the acceptance tests from their own forks** — they have to run them locally and hope CI passes after opening the PR.

This makes the same `e2e` job runnable from a fork, in two parts:

**1. Credentials** — thread `tps-action`'s existing inputs through from repository secrets/variables. They resolve as:

1. **`HCLOUD_TOKEN`** secret set → used directly (simplest; a token from a throwaway Hetzner project), else
2. contact TPS at **`tps-url`** (override via the **`TPS_URL`** variable; defaults to the public TPS) authenticating with a **`TPS_TOKEN`** secret if set, else GitHub OIDC.

`CERT_DOMAIN` is likewise made overridable via a variable (it was hardcoded), so certificate tests can point at a domain in the fork's own account. This matches how `hcloud-cloud-controller-manager`'s `test.yml` already threads `token: ${{ secrets.HCLOUD_TOKEN }}` into `tps-action`.

**2. A trigger a fork can fire** — add a `workflow_dispatch` trigger (and run `e2e` on it). A `pull_request` run from a fork executes in this repository's context, where fork secrets are unavailable — so the token fallback alone isn't reachable from a fork PR. `workflow_dispatch` lets a contributor run the suite **in their own fork's context**, where their fork secrets exist. The README documents the one-time setup (`HCLOUD_TOKEN` secret + `CERT_DOMAIN` variable) and the Actions → Run workflow step.

**Backward compatible:** on this repository none of the new secrets/variables are set, so every value falls back to its current default and the existing OIDC-against-default-TPS flow is unchanged. `workflow_dispatch` adds a manual trigger but changes nothing about the automatic `push`/`pull_request` runs.

## Testing

- `test.yml` validated as YAML; the `tps-action` inputs match the action's documented `token` / `tps-url` / `tps-token` interface, and each expression falls back to today's value when the secret/variable is unset.
- The `e2e` job's `if:` now also runs on `workflow_dispatch`; unit/lint/build are unaffected and still run on every pull request (including drafts).
- No behavioural change for the upstream repo (all new inputs resolve to empty/default; `workflow_dispatch` is opt-in).

### AI assistance

Developed with AI assistance (Claude Code, Anthropic — Opus 4.x). **AI's role:** drafted the workflow/README changes and verified the `tps-action` input semantics against the action's README and the sibling `hcloud-cloud-controller-manager` workflow. **My role:** identified the fork-can't-run-acctests gap, designed the token-fallback + `workflow_dispatch` approach, reviewed every line, and I take responsibility for it. Happy to adjust to whatever shape you prefer.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -78,6 +78,16 @@ jobs:
           go-version-file: go.mod
 
       - uses: hetznercloud/tps-action@dc6d67c5768ca43b840429b6c7409cda61cf4092 # main
+        with:
+          # Resolve credentials so the acceptance tests can also run from forks, where the
+          # default OIDC identity is not trusted by the public TPS:
+          #   - if HCLOUD_TOKEN is set, use it directly (skips TPS)
+          #   - otherwise contact TPS at tps-url (override with the TPS_URL variable; defaults
+          #     to the public TPS), authenticating with TPS_TOKEN if set, else GitHub OIDC.
+          # Upstream leaves all three unset, so the current OIDC-against-default-TPS flow is unchanged.
+          token: ${{ secrets.HCLOUD_TOKEN }}
+          tps-url: ${{ vars.TPS_URL || 'https://tps.hc-integrations.de' }}
+          tps-token: ${{ secrets.TPS_TOKEN }}
 
       - if: matrix.tool == 'opentofu'
         run: |
@@ -88,8 +98,9 @@ jobs:
       - run: go test -v -timeout=30m -parallel=8 -coverprofile=coverage.txt -coverpkg=./... ./...
         env:
           # Domain must be available in the account running the tests. This domain is
-          # available in the account running the public integration tests.
-          CERT_DOMAIN: hc-integrations-test.de
+          # available in the account running the public integration tests; forks can point
+          # the certificate tests at a domain in their own account via the CERT_DOMAIN variable.
+          CERT_DOMAIN: ${{ vars.CERT_DOMAIN || 'hc-integrations-test.de' }}
 
           TF_ACC: 1
           TF_LOG: DEBUG
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -112,8 +112,10 @@ jobs:
           TF_LOG: DEBUG
           TF_LOG_PATH_MASK: test-%s.log
 
-          TEST_DATACENTER: ${{ vars.TEST_DATACENTER }}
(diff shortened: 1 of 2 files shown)


nikolauspschuetz:
Note on CI: the `e2e` legs will go red on this PR's own `pull_request` run — that's the exact gap this change addresses. A fork `pull_request` executes in this repo's context without the fork's secrets, so `tps-action` still falls back to OIDC-against-TPS and is rejected. The new path is meant to be exercised via `workflow_dispatch` **in the fork** (with an `HCLOUD_TOKEN` secret set there), per the README section. On this repository the automatic OIDC flow is unchanged. Happy to demonstrate a green fork run if useful.

nikolauspschuetz:
Validated the fork path end-to-end ✅

I ran the acceptance tests from my fork via `workflow_dispatch` using an `HCLOUD_TOKEN` from my own Hetzner Cloud project: `tps-action` picked up the token and skipped TPS, and `TestAccSSHKeyResource` created and destroyed a real SSH key — green on **all five tool legs** (opentofu 1.10/1.11/1.12 + terraform 1.14/1.15) in ~2 minutes:
https://github.com/nikolauspschuetz/terraform-provider-hcloud/actions/runs/29383479930

Honest note: to keep that run free I used a fork-only branch that also adds a small `-run`/package selector (not part of this PR) so I could target a single free resource. The credential path *this* PR adds is what's exercised — the fork's own token driving the `e2e` job.

Also pushed in response to some self-review:
- Consolidated the fork instructions in the README into a single section.
- Defaulted `TEST_LOCATION` / `TEST_DATACENTER` (`hel1` / `hel1-dc2`) so a fork that leaves those variables unset doesn't export an empty value that overrides the Go-level default and breaks location-dependent tests.

One thing I left as-is: certificate tests still rely on a fork setting its own `CERT_DOMAIN` (otherwise they run against `hc-integrations-test.de`). I can make that default skip-on-unset for forks if you'd prefer — didn't want to touch the upstream default without your steer.

jooola:
Closed in favor of #1505

PR #1480, opened 2026-07-08
fix(load_balancer): lift delete protection when deleting the resource

### LINKED ISSUE #1206: [Bug]: delete_protection doesnt work as documented
### What happened?

The [delete protection docs](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs#delete-protection) state that resources with delete protection can be deleted by terraform and only prevent other api consumers (e.g. the cloud console) to delete the resource.

This is especially useful when using the hcloud-cloud-controller-manager and an existing LoadBalancer that was created with Terraform.

Terraform logs `load balancer deletion is protected` when trying to delete the load balancer resource.

I expect the load balancer to be deleted as documented.

```hcl
resource "hcloud_load_balancer" "ingress" {
  name               = "${var.cluster_name}-ingress"
  load_balancer_type = var.ingress.load_balancer.type
  location           = var.ingress.load_balancer.region
  delete_protection  = true # will prevent the hcloud-cloud-controller-manager to delete the lb so that terraform can delete it
}
```

nikolauspschuetz:
Closes #1206.

### Problem

The [delete-protection docs](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs#delete-protection) state that a protected resource can still be deleted by Terraform — protection only prevents *other* API consumers (e.g. the Cloud Console) from deleting it. But `hcloud_load_balancer` failed with `load balancer deletion is protected` instead of deleting.

As @apricote noted in the issue, only `hcloud_zone_rrset` currently lifts protection to make its changes; the other resources don't follow the documented behaviour.

### Fix

In `resourceLoadBalancerDelete`, if the Load Balancer has delete protection enabled, disable it (via the existing `setProtection` helper) before deleting — mirroring `hcloud_zone_rrset`. Adds `TestAccLoadBalancerResource_DeleteProtectionLifted`, which creates a Load Balancer with `delete_protection = true` and never disables it, so the framework's `CheckDestroy` must delete a still-protected resource (fails without this change).

### Scope

I've scoped this to **`hcloud_load_balancer`** — the resource in the issue (and the one that bites the `hcloud-cloud-controller-manager` use case). Your inventory in #1206 lists the same gap for `floating_ip`, `network`, `primary_ip`, `server`, `volume`, and `zone`. Happy to extend this PR to cover the rest in one coordinated change if you'd prefer that over per-resource PRs — just let me know.

### Testing

`go build ./...`, `go vet`, and `gofmt` are clean; the new acceptance test compiles and is skipped without `TF_ACC`. I haven't run it against a live account in this PR — glad for CI / a maintainer to run it, and I can also run it against my own throwaway project if useful.

### AI assistance
Implemented with Claude Code; I directed the change (reusing the in-repo `setProtection` helper and the `zone_rrset` pattern) and verified build/vet/test-compile locally.

### DIFF
--- a/internal/loadbalancer/resource.go
+++ b/internal/loadbalancer/resource.go
@@ -363,6 +363,15 @@ func resourceLoadBalancerDelete(ctx context.Context, d *schema.ResourceData, m a
 		return nil
 	}
 
+	// As documented, delete protection only prevents other API consumers from
+	// deleting the resource; Terraform lifts the protection so it can delete the
+	// Load Balancer it manages. (#1206)
+	if loadBalancer.Protection.Delete {
+		if err := setProtection(ctx, client, loadBalancer, false); err != nil {
+			return hcloudutil.ErrorToDiag(err)
+		}
+	}
+
 	if _, err := client.LoadBalancer.Delete(ctx, loadBalancer); err != nil {
 		if hcloud.IsError(err, hcloud.ErrorCodeNotFound) {
 			// loadBalancer has already been deleted
--- a/internal/loadbalancer/resource_test.go
+++ b/internal/loadbalancer/resource_test.go
@@ -248,3 +248,36 @@ func TestAccLoadBalancerResource_Protection(t *testing.T) {
 		},
 	})
 }
+
+// TestAccLoadBalancerResource_DeleteProtectionLifted is a regression test for
+// #1206: Terraform must be able to delete a Load Balancer that still has delete
+// protection enabled (protection only guards against other API consumers). The
+// configuration never disables protection, so the CheckDestroy step deletes a
+// still-protected Load Balancer, which fails without the fix.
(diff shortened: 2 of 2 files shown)


jooola:
Hey, I've added a comment to #1206, where I find it risky to change the behavior of the delete protection behavior within a minor release.

So this PR alone will not suffice to fix #1206.

nikolauspschuetz:
Thanks @jooola — that's fair. Silently changing delete-protection behavior in a
minor release is risky, and a provider-wide setting that lets users opt in while
preserving today's behavior is the safer path.

Happy to implement that. Before I rework this PR: do you have a preferred shape
for the setting — e.g. a `delete_protection` behavior attribute on the provider
block — and should it default to the current (existing) behavior? I'll align the
change in #1206 to whatever you and @apricote settle on.

PR #1501, opened 2026-07-24
feat: remove `datacenter` from primary ips and servers

jooola:
Mark the deprecated `datacenter` attribute for removal in the Server and Primary IP resources and data sources. Since the property was already removed from the Hetzner Cloud API, we do not consider this a breaking change.

See our [deprecation](https://docs.hetzner.cloud/changelog#2025-12-16-phasing-out-datacenters) and [removal](https://docs.hetzner.cloud/changelog#2026-07-01-removing-datacenters) changelog for more details.

---

The attribute was not removed from the state to reduce the impact of the change for the users. A state upgrade would prevent users to rollback their version to an older version. The attribute must be removed in a future major release.


jooola:
Next step is tracked here: https://github.com/hetznercloud/terraform-provider-hcloud/issues/1502

apricote (approved):

PR #1512, opened 2026-07-27
feat(load_balancer_type): add deprecation details and warnings

jooola:
- Add the deprecation details to the load balancer type schemas
- Add deprecation warning diagnostic when the load balancer type is deprecation or unavailable
- Improve our deprecation util package to not store empty strings instead of null for new attributes (not migrated from the sdk).
- Add deprecation warning diagnostic when the server type is deprecation or unavailable.

```rp-commits
feat(load_balancer_type): add deprecation details and warnings (#1512)
feat(server_type): add deprecation warnings (#1512)
```

### DIFF
--- a/internal/servertype/data_source.go
+++ b/internal/servertype/data_source.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 
 	"github.com/hetznercloud/hcloud-go/v2/hcloud"
+	"github.com/hetznercloud/hcloud-go/v2/hcloud/exp/deprecationutil"
 	"github.com/hetznercloud/terraform-provider-hcloud/internal/deprecation"
 	"github.com/hetznercloud/terraform-provider-hcloud/internal/util"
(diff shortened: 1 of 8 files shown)


internal/servertype/data_source.go
```diff
@@ -329,6 +330,14 @@ func (d *dataSource) Read(ctx context.Context, req datasource.ReadRequest, resp
...
 	}
 
+	if message, unavailable := deprecationutil.ServerTypeMessage(result, ""); message != "" {
+		if unavailable {
+			resp.Diagnostics.AddWarning("Server Type unavailable", message+".")
+		} else {
+			resp.Diagnostics.AddWarning("Server Type deprecated", message+".")
+		}
+	}
+
```
lukasmetzner, line 340: Might be worth a second release note entry.

lukasmetzner (approved):
Looking good 🚀

internal/servertype/data_source.go
jooola: Noted

PR #1514, opened 2026-08-13
ci: allow checkout arbitrary refs

jooola:
Allow to checkout any reference provided as input. This is useful for running our integrations tests against pull requests from forks.

After verifying that the pull request is sane, you can trigger the e2e tests using the following command:
```sh
# Where 1468 is the PR number and origin is the git remote for this project
REF="$(git ls-remote origin pull/1468/head | head -1 | cut -f 1)"
# Where test.yml is the workflow id / filename
gh workflow run -f "ref=$REF" test.yml
```


lukasmetzner (approved):
Nice 🚀

PR #1524, opened 2026-09-07
feat(storage_box): support write-only password attributes

### LINKED ISSUE #1299: [Feature]: Support write-only password attributes in storage-boxes subaccounts

Hello :) 

Recently Opentofu support ephemeral ressources (available in terraform since 1.10), so I try to make a subaccounts in my storage box for each vm I have create.

Because we can add an SSH key on a storage box subaccounts via API, I generate a random password, store it into my vault, then I run a playbook to install and configure borgmatic (using password stored in vault).

```tf
data "hcloud_storage_box" "backup" {
  name = "backup"
}

ephemeral "random_password" "backup_password" {
  for_each         = var.incus_servers

  length           = 128
  special          = true
  override_special = "ÄÖÜäöü^°!§$%/()=?+#-.,;:~*@{}_&"
}

resource "vault_kv_secret_v2" "db_root" {
  for_each         = var.incus_servers

  mount        = var.vault_kv_engine
  name 			     = "storage-boxes/${each.key}"
  data_json_wo                  = jsonencode(
    {
      password       = ephemeral.random_password.backup_password[each.key].result
    }
  )
}

resource "hcloud_storage_box_subaccount" "team_badger" {
  for_each         = var.incus_servers

  storage_box_id = data.hcloud_storage_box.backup.id
  description = each.key

  home_directory = "borg/${each.key}"
  password       = ephemeral.random_password.backup_password[each.key].result

  access_settings = {
    reachable_externally = true
    ssh_enabled = true
  }

  labels = {
    env  = "production"
    tool = "terraform"
  }
}
```

But terraform send this error

> Attribute ".password" is referencing an ephemeral value but ephemeral values can be referenced only by other ephemeral attributes or by write-only ones.

Vault provider (for exemple) add an attribute name `data_json_wo` instead of `data_json` to write only and don't log anywhere the password.

So, I think for a fully and securelly automation, a `password_wo` attribute can be amazing !

Thank :) 
Have a nice day

jfilter:
Closes #1299

Adds `password_wo` and `password_wo_version` to `hcloud_storage_box` and `hcloud_storage_box_subaccount`, so the Storage Box password never reaches the Terraform state. `password` becomes optional; exactly one of `password` and `password_wo` must be set (`ExactlyOneOf`), and `password_wo` requires `password_wo_version`.

A write-only value is null in plan and state, so it cannot be compared against either. A password reset is therefore triggered by incrementing `password_wo_version` — the same model the HashiCorp providers use (e.g. `aws_db_instance.password_wo_version`). Moving an existing configuration from `password` to `password_wo` drops the stored password from the state on the next apply.

Write-only arguments need Terraform 1.11 / OpenTofu 1.11 or later; the new acceptance tests skip below that, so the OpenTofu 1.10.x job in the matrix is unaffected.

- Schema, model and Create/Update paths for both resources
- Docs, examples and test templates updated (`go generate` is clean)
- Acceptance tests: create with `password_wo`, unchanged plan on same version, reset on incremented version, `password`/`password_wo` never written to state


internal/storagebox/resource.go
```diff
@@ -106,9 +109,28 @@ See the [Storage Box API documentation](https://docs.hetzner.cloud/reference/het
...
+				stringvalidator.AlsoRequires(path.MatchRoot("password_wo_version")),
+			},
+		},
+		"password_wo_version": schema.Int64Attribute{
+			MarkdownDescription: "Version of `password_wo`. The value of `password_wo` cannot be compared against the API or the state, so a password change is triggered by incrementing this instead.",
+			Optional:            true,
+			Validators: []validator.Int64{
+				int64validator.AlsoRequires(path.MatchRoot("password_wo")),
+			},
 		},
```
jooola, line 124: From the usability perspective, do you think a string gives more freedom to the users, and they can use any values they like? For example a date? Or a number, but as string? What is the benefit of having a integer?

internal/storagebox/resource.go
```diff
@@ -233,18 +273,30 @@ func (m *resourceModel) ToTerraform(ctx context.Context) (types.Object, diag.Dia
...
 	if resp.Diagnostics.HasError() {
 		return
 	}
 
+	password, ok := data.password(&config)
+	if !ok {
+		resp.Diagnostics.AddError(
+			"Missing password",
+			"Exactly one of `password` and `password_wo` must be set, and neither carries a value.",
+		)
```
jooola: This should be validated using the resource validation features (not at the attribute level, but at the resource level).

jooola (commented):
Hey, thanks for the contribution.

I did a first review, with some comments.

internal/storagebox/resource.go
jfilter: I just looked at how others are solving this, and it seems that most providers use an integer for this (AWS, Azure). From the usability perspective, I think it's easier to just increment it here.

internal/storagebox/resource.go
jfilter: Thanks, fixed in a new commit.

PR #1527, opened 2026-09-08
feat(server): always read `public_net` config into state

jooola:
This ensure that when importing a server, the plan does not propose a public network update, forcing a server power cycle.


lukasmetzner (approved):
Is there an import e2e test we can add, or extend to cover this case?

jooola:
Ah my bad, I am still cooking on this one.

jooola:
This change it too "big", we do not have a grasp of the repercussion of such a change. Closing in favor of #1534

PR #1534, opened 2026-09-14
feat(server): save public net state during import

jooola:
To prevent a power cycle during the first update after an import, we must write the actual state of the server in the terraform state so that no diff is generated.

With this change, the server creation using terraform and a server import generate a similar state (with the public_net attribute populated).

The public net update logic is currently not smart enough and not idempotent, to skip any unneeded changes. It will likely always power off the server, not change anything, and power it back on.


jooola:
This solution does not cover the case when user does not have a public_net configured.

